PR #21, opened 2018-06-22
Fix CVE-2018-3760

danielbreves:
Introduced in https://github.com/zendesk/volunteer/pull/244/files#diff-1d98f0039bbc6df92adee882ed99d993R30

https://blog.heroku.com/rails-asset-pipeline-vulnerability

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -23,7 +23,7 @@ Rails.application.configure do
   config.assets.css_compressor = :sass
 
   # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = true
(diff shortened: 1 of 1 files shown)


craig-day (changes requested):
I’m pretty sure this will break things for our deployment. We don’t ship our assets to a cdn or anything. Isn’t a bump to sprockets enough?

danielbreves:
@craig-day that option is for [live compilation](http://guides.rubyonrails.org/asset_pipeline.html#live-compilation) only, no? Are we not running `rake assets:precompile` before deploying?

craig-day (approved):

PR #33, opened 2018-07-18
Make sure PostgreSQL is listed before bundling

waferbaby:
Without install PostgreSQL first, bundling may fail because the `pg_config` command is unavailable.


nchieng (approved):

fvzzy (approved):

craig-day:
this is very similar to #38

PR #35, opened 2018-07-18
add redux devtools for easier state debugging

nchieng:
Prefer using the browser extension to inspect state than following the output on console

### DIFF
--- a/app/javascript/store/configureStore.dev.js
+++ b/app/javascript/store/configureStore.dev.js
@@ -4,10 +4,11 @@ import createLogger from 'redux-logger'
 import createRootReducer from '../reducers'
 
 export default function configureStore(apolloClient, initialState = {}) {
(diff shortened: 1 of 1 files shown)


craig-day (approved):
Assuming this can't blowup with some undefined error, looks good to me

nchieng:
I have not experienced any issues on Central Admin.

And it is restricted to dev.

I've just tested this successfully with a few different browsers with and without the Redux Devtool extension.

craig-day:
@nchieng do you still want to merge this?

nchieng:
I don't see any harm in doing so.
However, given how old this PR is, I think it is better to close it without merging.
If there are other contributors that find that they are missing this, it would be easier to dredge this back up.

craig-day:
Okay, closing for now. Thanks!

PR #37, opened 2018-07-18
Ensure end time is after start time

yifeiwu:
Ensure end time is after start time for an event.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -16,6 +16,7 @@ class Event < ApplicationRecord
   validates :capacity, numericality: { only_integer: true }
 
   validate :max_capacity
+  validate :time_sequentiality
 
   scope :before,     ->(time) { where("starts_at < ?", time) }
   scope :after,      ->(time) { where("starts_at > ?", time) }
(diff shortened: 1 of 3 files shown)


xuebingli (approved):
👍 assuming that time only flows in one direction and that very direction is forward :trollface:

xuebingli (approved):
👍 assuming that time only flows in one direction and that very direction is forward :trollface:

yifeiwu:
Go home Dr Strange, you're drunk

nchieng (approved):

ragurney (approved):

craig-day (changes requested):
please add a server side validation too

yifeiwu:
@craig-day added server side tests

xuebingli:
@craig-day 💇‍♂️

app/models/event.rb
```diff
@@ -44,9 +45,11 @@ def remove_user!(user)
   private
 
   def max_capacity
-    if capacity && users.size > capacity
-      errors.add(:users, "total can't be above capacity")
-    end
+    errors.add(:users, "total can't be above capacity") if capacity && users.size > capacity
```
craig-day: don't love this style change, the line is pretty long now.

app/models/event.rb
```diff
@@ -44,9 +45,11 @@ def remove_user!(user)
...
+  end
+
+  def time_sequentiality
+    errors.add(:ends_at, "should not end before it starts") if starts_at && ends_at && ends_at <= starts_at
```
craig-day: I would prefer an if block, not a trailing if when its this long.

craig-day (approved):

Issue #36, opened 2018-07-18
Individual events: event duration should have a sensible min/max range


fvzzy:
Can currently set these to negative values with no upper bound
<a href="https://cl.ly/1h2c3F0c3k37" target="_blank"><img src="https://d3vv6lp55qjaqc.cloudfront.net/items/1R0J2y3j151F303Y2W0D/Image%202018-07-18%20at%202.37.50%20pm.png" style="display: block;height: auto;width: 100%;"/></a>

PR #41, opened 2018-07-18
Add basic validation for the Individual Events duration field

fvzzy:
Addresses https://github.com/zendesk/volunteer_portal/issues/36

### DIFF
--- a/app/javascript/pages/IndividualEvents/index.js
+++ b/app/javascript/pages/IndividualEvents/index.js
@@ -189,6 +189,7 @@ const CreateEditDialog = ({ offices, eventTypes, organizations, onCancel, popove
               name="duration"
               label="Duration (minutes)"
               type="number"
+              normalize={value => (value >= 0 ? value : 0)}
               component={renderField}
               className={s.input}
             />
--- a/app/javascript/pages/IndividualEvents/index.js
+++ b/app/javascript/pages/IndividualEvents/index.js
@@ -96,6 +96,8 @@ const validate = values => {
   }
   if (!values.duration) {
     errors.duration = 'is required'
+  } else if (values.duration < 24 * 60) {
+    errors.duration = 'must be less than or equal to 24 hours'
   }
   if (!values.eventType) {
     errors.eventType = {}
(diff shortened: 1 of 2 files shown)


danielbreves (approved):

app/javascript/pages/IndividualEvents/index.js
```diff
@@ -97,6 +97,9 @@ const validate = values => {
   if (!values.duration) {
     errors.duration = 'is required'
   }
+  if (values.duration > 8 * 60) {
```
craig-day: where did the 8 hours come from? Before we merge this I would like to confirm with everyone that 8 hours is the max.

craig-day (changes requested):
please also add a server side validation for this

xuebingli:
@craig-day feedback addressed 💇‍♂️

app/javascript/pages/IndividualEvents/index.js
xuebingli: @azaiszen from CSR Committee suggests that 24 hours is a reasonable limit.

app/models/individual_event.rb
```diff
@@ -11,7 +11,10 @@ class IndividualEvent < ApplicationRecord
   belongs_to :office
 
   validates :user, :organization, :event_type, :office,
-    :description, :date, :duration, presence: true
+            :description, :date, presence: true
+  validates :duration, numericality: { only_integer: true,
+                                       greater_than: 0,
+                                       less_than_or_equal_to: 1440 }
```
craig-day: lets take advantage of rails here and make this more readable

```rb
less_than_or_equal_to: 24.hours
```

craig-day (approved):
Looks good after the time helper addition

app/models/individual_event.rb
xuebingli: I'm afraid this doesn't work as expected. `duration` on `IndividualEvent` is an integer representing minutes.

```
irb(main):018:0> 61 < 1.hour
=> true
```

craig-day: Okay

PR #43, opened 2018-07-18
Allow Admins to remove users from Events

brianac:
Scoped this down to simply removing users for now.

deleteusers

- Note the spacing of the Remove link looks off center vertically because none of this demo data has subtags
- I was trying to get the Remove button to position itself to the right of the flex container but didn't have any success.

### DIFF
--- a/app/javascript/components/NamedAvatar.js
+++ b/app/javascript/components/NamedAvatar.js
@@ -16,6 +16,10 @@ const styles = {
   name: {
     fontWeight: 600,
   },
+  remove: {
+    marginLeft: '20px',
+    cursor: 'pointer',
+  },
   subtitle: {},
 }
 
@@ -26,6 +30,11 @@ const NamedAvatar = props => (
       <span style={styles.name}>{props.name}</span>
       <span style={styles.subtitle}>{props.subtitle || '\u00a0'}</span>
     </div>
+    {props.showRemove && (
+      <a style={styles.remove} onClick={props.onRemove}>
+        Remove
+      </a>
+    )}
   </div>
 )
 
--- a/app/javascript/components/NamedAvatar.js
+++ b/app/javascript/components/NamedAvatar.js
@@ -1,5 +1,8 @@
 import React from 'react'
 import Avatar from './Avatar'
+import Dialog from 'material-ui/Dialog'
+import { connect } from 'react-redux'
+import { togglePopover } from 'actions'
 
 const styles = {
   container: {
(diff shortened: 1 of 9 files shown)


app/javascript/pages/admin/Events/edit.js
```diff
@@ -8,21 +8,50 @@ import { graphQLError } from 'actions'
...
+      <EventForm
+        event={event}
+        eventTypes={eventTypes}
+        offices={offices}
+        organizations={organizations}
+        onSubmit={updateEvent}
+      />
+      Signed up Users
+      {event.users.map((user, i) => (
+        <div key={i}>
```
craig-day: can we maybe extract and reuse the user list that is on the regular event details page? Then we could just pass in a prop if someone is an admin to have the add/remove controls added.

app/javascript/pages/admin/Events/queries/index.gql
```diff
@@ -1,12 +1,16 @@
...
 #import "fragments/OfficeEntry.gql"
+#import "fragments/UserEntry.gql"
 
 query EventsQuery {
   events {
     ...EventEntry
     office {
       ...OfficeEntry
     }
+    users {
```
craig-day: I'm not crazy about doing this on the event collection... thats a lot of data to shove into the response. Where is this used?

craig-day (changes requested):
I would prefer we re-use the users-for-event component page if possible so we get the grid formatting

app/javascript/pages/admin/Events/queries/index.gql
brianac: Oh you're right this was a mistake. Will remove it

brianac:
@craig-day 💇

brianac:
@craig-day bump

Mehul-Porwal:
@brianac Added a commit on top of your branch to support confirmation dialog before removing user from event.

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +19,78 @@ const styles = {
...
+    fontSize: '16px',
+  },
 }
 
-const NamedAvatar = props => (
+const dialogActions = (togglePopover, onRemove) => [
+  <a
+    style={styles.dialogActionsLink}
+    onClick={() => {
+      togglePopover('removeUserFromEvent', false)
```
thebritican: [We're being truthy](https://github.com/zendesk/volunteer_portal/blob/b87c36e787789117f16e36220aa5186a7a50430d/app/javascript/actions/index.js#L29) in the `togglePopover` function definition with the second argument (`data`).

But, i think the second argument should either be provided with `null` or excluded entirely since it's only meant to save retain popover specific data in the state. I should document that 😓. 

So `togglePopover('removeUserFromEvent')` is fine

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +19,78 @@ const styles = {
...
+      <a
+        style={styles.remove}
+        onClick={() => {
+          togglePopover('removeUserFromEvent', true)
+        }}
+      >
+        Remove
+      </a>
+    )}
+    {!!popover ? (
```
thebritican: nit: this is just fine for people who know about JS trickery 😈 but since we're already using Ramda, how would you feel about inverting this with a more explicit function?

```js
R.isNil(popover) ? null : <Dialog />.
```

It's not the best handling of truthy, but at least it handles `null` _and_ `undefined`

thebritican: (i also realize i've written `!!` a ton... so it would take a while to replace it everywhere) do whatever you like!

gregziegan (approved):

app/javascript/components/NamedAvatar.js
craig-day: I would prefer the `R.isNil` check. We have done a good job of not using `!!` and truthy/falsy stuff in this project so far.

craig-day (commented):
Looks pretty good. Should there be a confirmation?

craig-day (approved):
I would prefer the switch to `R.isNil` instead of `!!` still.

xuebingli (changes requested):
❗️**DO NOT MERGE** ❗️ 
This PR currently has the known issue of removing the wrong user. I'll put up a fix if I've got the time.

xuebingli:
This issue is reproducible.

xuebingli:
Superseded by #87

Issue #49, opened 2018-08-01
Enable listing of events in most recent order


jburdett:
At the moment events are displayed typically from first one created down to the last.  
So if you want to view the event you just created you have to scroll all the way to the bottom.
If you have a lot of events that's a lot of scrolling.
Preferably you should be able to indicate some sort of sorting mechanism.
Ideally you can sort by any column.

PR #50, opened 2018-08-01
Sort events by descending

jburdett:
Makes default sorting for the admin page listing events show the latest event first.
This is a first step, there is more that needs to be sorted.
To test it out navigate to `http://localhost:5000/portal/admin/events`
Newest events should be at the top

For https://github.com/zendesk/volunteer_portal/issues/49

### DIFF
--- a/app/graph/resolvers/event_resolver.rb
+++ b/app/graph/resolvers/event_resolver.rb
@@ -1,5 +1,10 @@
 module EventResolver
   class << self
+    def all(_, args, _)
+      scope = Event.all
+      scope = scope_with_sort_by(scope, args[:sortBy])
+    end
+
     def create(_, args, context)
       attrs = args[:input].to_h
 
@@ -39,6 +44,15 @@ module EventResolver
       event.capacity = attrs['capacity']
       event.starts_at = Time.parse(attrs['startsAt'])
       event.ends_at = Time.parse(attrs['endsAt'])
+    end
+
+    def scope_with_sort_by(scope, sort_by)
+      return scope if sort_by.nil?
+
+      # Makes an ORDER BY string if it follows format of "[column_name]_[order]"
+      # For example: "STARTS_AT_DESC" -> "starts_at DESC"
+      query_string = [sort_by.rpartition('_').first.downcase, sort_by.rpartition('_').last].join ' '
+      scope.order(query_string)
     end
   end
 end
--- a/app/graph/resolvers/event_resolver.rb
+++ b/app/graph/resolvers/event_resolver.rb
@@ -64,10 +64,12 @@ module EventResolver
 
     def scope_with_sort_by(scope, sort_by)
       return scope if sort_by.nil?
-
-      # Makes an ORDER BY string if it follows format of "[column_name]_[order]"
-      # For example: "STARTS_AT_DESC" -> "starts_at DESC"
-      query_string = [sort_by.rpartition('_').first.downcase, sort_by.rpartition('_').last].join ' '
+      query_string = case sort_by
+      when STARTS_AT_DESC
+        'starts_at DESC'
+      when STARTS_AT_ASC
+        'starts_at ASC'
+      end
       scope.order(query_string)
     end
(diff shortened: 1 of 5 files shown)


app/graph/resolvers/event_resolver.rb
```diff
@@ -40,5 +61,14 @@ def update_fields(event, attrs)
       event.starts_at = Time.parse(attrs['startsAt'])
       event.ends_at = Time.parse(attrs['endsAt'])
     end
+
+    def scope_with_sort_by(scope, sort_by)
+      return scope if sort_by.nil?
+
+      # Makes an ORDER BY string if it follows format of "[column_name]_[order]"
+      # For example: "STARTS_AT_DESC" -> "starts_at DESC"
+      query_string = [sort_by.rpartition('_').first.downcase, sort_by.rpartition('_').last].join ' '
```
xuebingli: Since sort_by is an enum and have only two values, why not just use `case`?

app/graph/resolvers/event_resolver.rb
```diff
@@ -1,5 +1,26 @@
...
+      else
+        args[:officeId]
+      end
+
+      events = Event.all
+      events = events.for_office(office_id)          if office_id
+      events = events.before(Time.at(args[:before])) if args[:before]
+      events = events.after(Time.at(args[:after]))   if args[:after]
+      events = scope_with_sort_by(events, args[:sortBy])
+      events
```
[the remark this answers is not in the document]
liulikun: You're right. `all` is a scope indeed. 👍

test/graph/resolvers/event_resolver_test.rb
```diff
@@ -72,4 +72,98 @@
...
+        starts_at: event3_start,
+        ends_at: event3_start + 1.hour,
+        capacity: 10,
+        location: 'somewhere',
+        office: other_office
+      )
+    end
+    let(:event1_start) { Time.new(2018, 1, 1, 1, 0) }
+    let(:event2_start) { Time.new(2018, 1, 1, 3, 0) }
+    let(:event3_start) { Time.new(2018, 1, 1, 2, 0) }
```
danielbreves: I'd find this test a little easier to read if 1, 2, and 3 was the actual start time order. Then unsorted could be `[event2, event1, event3]`, asc would be `[event1, event2, event3]`, desc would be `[event3, event2, event1]`.

danielbreves (approved):

app/javascript/pages/admin/Events/index.js
```diff
@@ -19,6 +19,8 @@ import s from './main.css'
 
 import 'style-loader!css-loader!react-table/react-table.css'
 
+const eventsSort = 'STARTS_AT_ASC'
```
danielbreves: Shouldn't this be `STARTS_AT_DESC`?

danielbreves: Shouldn't this be STARTS_AT_DESC?

danielbreves (changes requested):

app/javascript/pages/admin/Events/index.js
jburdett: 🤦‍♂️

danielbreves: It seems like it would also be possible to resize all headings client-side by removing `sortable: false` from [here](https://github.com/zendesk/volunteer_portal/blob/7b4d8b1d85bba0fb65a0eac1c76219b4c2ac4dfa/app/javascript/pages/admin/Events/index.js#L51). Not sure why it was marked as `false` in the first place. @craig-day @roseleaf may know?

danielbreves (approved):

xuebingli (approved):

app/javascript/pages/admin/Events/index.js
jburdett: @craig-day  or @roseleaf I would also like to know the answer to @danielbreves 's question ☝️

PR #62, opened 2018-08-29
Display dashboard filtered by current office by default

danielbreves:
Display dashboard filtered by current office by default


app/javascript/pages/Dashboard/index.js
```diff
@@ -141,12 +130,12 @@ const LeaderboardContainer = ({
         <div className={s.topVolunteers}>Top Volunteers</div>
         <Filter
           collection={offices}
-          value={dashboardOfficeFilter.value}
+          value={dashboardOfficeFilter.value === 'current' ? currentUser.office.id : dashboardOfficeFilter.value}
           onChange={(_event, _index, value) => changeDashboardOfficeFilter(value)}
           itemValueProp="name"
         />
       </div>
-      {topUsers(dashboardOfficeFilter, users).map((user, i) => (
```
danielbreves: This doesn't seem to be needed since it's already sorted by hours in the server here: https://github.com/zendesk/volunteer_portal/blob/master/app/graph/resolvers/user_resolver.rb#L92

app/javascript/pages/Dashboard/index.js
```diff
@@ -141,12 +130,12 @@ const LeaderboardContainer = ({
...
+      {users.map((user, i) => (
         <div className={s.leaderboardUser} key={`user-${i}`}>
```
thebritican: slightly out of scope of this PR but we should just use `user.id` here, no? no need for the `index` argument then

gregziegan (approved):

Issue #55, opened 2018-08-29
Make admin events page sortable by heading


danielbreves:

xuebingli:
Addressed by #64

PR #64, opened 2018-08-29
Sort events by clicking headings

jburdett:
Solves https://github.com/zendesk/volunteer_portal/issues/55
Uses the default sorting behaviour that is packaged with ReactTable

### DIFF
--- a/app/javascript/pages/admin/Events/index.js
+++ b/app/javascript/pages/admin/Events/index.js
@@ -38,6 +38,7 @@ const columns = deleteEvent => [
   {
     Header: 'Office',
     accessor: 'office.name',
+    minWidth: 70,
     sortable: true,
   },
   {
--- a/app/javascript/pages/admin/Events/index.js
+++ b/app/javascript/pages/admin/Events/index.js
@@ -34,27 +34,28 @@ const columns = deleteEvent => [
   {
     Header: 'Title',
     accessor: 'title',
+    sortable: true,
   },
   {
     Header: 'Office',
     accessor: 'office.name',
-    sortable: false,
+    sortable: true,
   },
   {
     Header: 'Description',
     accessor: 'description',
-    sortable: false,
+    sortable: true,
   },
   {
     Header: 'Start',
     accessor: 'startsAt',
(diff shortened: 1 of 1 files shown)


danielbreves (approved):

app/javascript/pages/admin/Events/index.js
```diff
@@ -34,27 +34,28 @@ const columns = deleteEvent => [
...
     Header: 'Office',
     accessor: 'office.name',
-    sortable: false,
+    sortable: true,
   },
   {
     Header: 'Description',
     accessor: 'description',
-    sortable: false,
+    sortable: true,
```
xuebingli: I'm curious if admins need to sort by description.

app/javascript/pages/admin/Events/index.js
```diff
@@ -34,27 +34,28 @@ const columns = deleteEvent => [
...
   },
   {
     Header: 'Start',
     accessor: 'startsAt',
-    sortable: false,
+    sortable: true,
```
xuebingli: Is ReactTable smart enough to recognize this column to be date & time and apply sorting accordingly, instead of sorting alphabetically?

jburdett: Good point I will go check

danielbreves:
I'm seeing this in Chrome:
screen shot 2018-09-05 at 5 43 02 pm

Can we set `outline-style: none;` on the headers?

danielbreves:
Also, it seems the column resizer is overflowing the next column, which makes the start of the headings not clickable:
screen shot 2018-09-05 at 5 50 04 pm

danielbreves:
Added column borders so the resizer feels a bit more user-friendly.

screen shot 2018-09-12 at 4 47 33 pm

app/javascript/pages/admin/Events/index.js
danielbreves: Looks like it:

screen shot 2018-09-12 at 4 52 05 pm

gregziegan (approved):

Issue #68, opened 2018-09-05
Cancel button on "Create Event" looks like it is disabled


jpandl19:
![image](https://user-images.githubusercontent.com/11040668/45125167-20143600-b122-11e8-9aa1-fff1607fd34e.png)

The gray color makes it seem like the "Cancel" button is disabled. Let's make it red or something.

PR #84, opened 2018-10-10
Modify cancel button color on Create Event form

keitheous:
We modified the color of the Cancel Button on Create and Edit Individual Event Page, according to Issue #68. For consistency, we went with the color of * Required (same row - far left).

<details>
<summary>Before</summary>
before
</details>

<details>
<summary>After</summary>
after
</details>


keitheous:
@yifeiwu Sure, we can do that!

craig-day (approved):

PR #87, opened 2018-10-10
Allow admin to remove user from event

xuebingli:
@zendesk/volunteer-eng

This PR allows an admin to removes users from event and supersedes #43. Built on top of #43, this PR includes refactoring and two bug fixes that

- Fix bug with user avatar rendering
- Fix bug that a wrong user is removed


xuebingli:
cc @craig-day @thebritican

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +20,79 @@ const styles = {
...
+          togglePopover('removeUserFromEvent', { userId })
+        }}
+      >
+        Remove
+      </a>
+    )}
+    {present(popover) &&
+      popover.data.userId === userId && (
+        <Dialog
+          title={'Remove User ' + name + ' from Event'}
```
craig-day: can use template strings here 

```jsx
title={`Remove User ${name} from Event`}
```

I would also maybe just drop the word "User" from the thing so it reads "Remove hello from Event"

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +20,79 @@ const styles = {
   name: {
     fontWeight: 600,
   },
-  subtitle: {},
+  remove: {
+    marginLeft: '20px',
```
craig-day: `px` is the default. You can just use the integer 20 here.

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +20,79 @@ const styles = {
...
+    cursor: 'pointer',
+  },
+  dialogActionsLink: {
+    padding: '15px',
+    cursor: 'pointer',
+    fontSize: '16px',
```
craig-day: same as above

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +20,79 @@ const styles = {
...
+    )}
+    {present(popover) &&
+      popover.data.userId === userId && (
+        <Dialog
+          title={'Remove User ' + name + ' from Event'}
+          actions={dialogActions(togglePopover, onRemove)}
+          open
+          onRequestClose={() => {
+            togglePopover('removeUserFromEvent')
+          }}
```
craig-day: any reason for the extra braces here? Can't it just be 

```jsx
onRequestClose={() => togglePopover()}
```

app/javascript/components/NamedAvatar.js
```diff
@@ -16,17 +20,79 @@ const styles = {
...
+          actionsContainerStyle={{ paddingBottom: 20, textAlign: 'center' }}
```
craig-day: can we move these styles to the `styles` hash for the file?

app/javascript/pages/admin/Admin/index.js
```diff
@@ -87,24 +87,15 @@ const OfficeFilter = ({ adminOfficeFilter, changeAdminOfficeFilter, offices }) =
...
 
   render() {
-    const {
-      routing,
-      children,
-      adminOfficeFilter,
-      changeAdminOfficeFilter,
-      data: { currentUser, offices },
-    } = this.props
+    const { routing, children, adminOfficeFilter, changeAdminOfficeFilter, data: { currentUser, offices } } = this.props
```
craig-day: did prettier single line this? The line seems pretty long.

app/javascript/pages/admin/Events/edit.js
```diff
@@ -11,15 +11,23 @@ import Loading from 'components/LoadingIcon'
 
 import EventQuery from './queries/show.gql'
 import UpdateEventMutation from './mutations/update.gql'
+import DestroySignupMutation from 'mutations/DestroySignupMutation.gql'
```
craig-day: does this need to be a relative like the other ones?

craig-day (changes requested):
Now that we are using the `present` helper in so many places, should we call it `isPresent`?

app/javascript/pages/admin/Admin/index.js
xuebingli: Yes, prettier did this. I could commit with `--no-verify` if this line is annoying and/or put up a separate PR to decrease the max line length setting of the linter.

xuebingli:
@craig-day bump

app/javascript/pages/admin/Admin/index.js
craig-day: nah, I trust our prettier config. It was just surprised.

craig-day (changes requested):
All of this code looks good, but lets add a check to ensure that the user executing the `destroySignup` mutation is either destroying their own signup, or is an admin. 

https://github.com/zendesk/volunteer_portal/blob/593c7b1309e916d1379e5a4ec53a375477af3ed9/app/graph/types/mutation_graph_type.rb#L23-L27

You should have the current user from the context, which is the third argument to the resolver.

xuebingli:
@craig-day thanks for reviewing the PR. I've addressed the comment.

craig-day (approved):

PR #96, opened 2018-10-24
added missing dependency `moment`

samzx:
Creating new event fails to load due to missing dependency. Adding dependency fixes the error.
@zendesk/volunteer
@liulikun


xuebingli (approved):

xuebingli:
I wonder if [snapshot testing](https://jestjs.io/docs/en/snapshot-testing) can help us prevent similar bugs in the future.

PR #102, opened 2018-11-07
[Fix] #85 Volunteer activity bar in calendar does not refresh

### LINKED ISSUE #85: Volunteer activity bar in calendar does not refreshes
When I clicked on a volunteering activity to join a volunteering activity, a pop up box appears. I clicked "sign up" and and my face turns up on the list - hooray! 

Then I clicked out of the pop up box, and realised the activity bar on the calendar remains the same. Still doesn't update until I hit the refresh button. Surely it should update immediately?

samzx:
When a user signs up to an event, it doesn't update the capacity filled bar. See #85 

## Description
This PR addresses #85, so that users get instant visual feedback from the event capacity bar when they sign up or leave an event. 

closes #85 

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/85)

## Implementation notes
The reason it does not update the horizontal bar beforehand is because the mutate query is handled by the floating event popup (where users sign up), which is a seperate component  to the horizontal  bar. They don't share this data, so it will not be updated.

To fix this, Apollo client provides an [update option](https://www.apollographql.com/docs/react/essentials/mutations.html) which allows us to update the cache after a mutation is called.

In this field, we update the cache, combining the mutation response with the current cache values. _There can be some cleaner code in combining these two data structures, maybe using that Ramda library?_ 🤷‍♀️

## CCs
@zendesk/volunteer

## Risks (if any)
* low, risk in the local cache

### DIFF
--- a/app/javascript/pages/Calendar/index.js
+++ b/app/javascript/pages/Calendar/index.js
@@ -140,22 +140,26 @@ const mapStateToProps = (state, ownProps) => {
   }
 }
 
+const momentAfter = Number(
+  moment()
+    .startOf('month')
+    .subtract(1, 'months')
+    .format('X')
+)
+
+const momentBefore = Number(
+  moment()
+    .endOf('month')
+    .add(1, 'months')
+    .format('X')
+)
+
 const withData = compose(
   graphql(EventsQuery, {
     options: ({ filters: { officeFilter: { value: officeId } } }) => ({
       variables: {
-        after: Number(
-          moment()
-            .startOf('month')
-            .subtract(1, 'months')
-            .format('X')
-        ),
-        before: Number(
-          moment()
-            .endOf('month')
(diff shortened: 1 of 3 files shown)


app/javascript/pages/Calendar/index.js
```diff
@@ -192,6 +219,29 @@ const withData = compose(
               }),
             },
           },
+          update: (cache, { data: { destroySignup: { event: eventChange } } }) => {
```
danielbreves: Could we reuse the callback above? Looks like only the parameter changes?

samzx: Good point, I felt like I forgot something 😅

danielbreves (approved):

keitheous:
nice work @samzx !

PR #106, opened 2018-11-19
Display offices alphabetically

### LINKED ISSUE #78: office list - sorting
should the office list be sorted alphabetically? Would make it much easier to find your office

blakegong:
## Description

Display offices alphabetically in all the related dropdowns.

## References

Closes #78.

## CCs

@rshokrizadeh @dineshsaravanan @shiyunl 

## Risks

Low, `offices` query related features might be affected.

### DIFF
--- a/app/graph/resolvers/office_resolver.rb
+++ b/app/graph/resolvers/office_resolver.rb
@@ -1,5 +1,16 @@
 module OfficeResolver
+  ALPHABETICAL_DESC = 'ALPHABETICAL_DESC'.freeze
+  ALPHABETICAL_ASC = 'ALPHABETICAL_ASC'.freeze
+
   class << self
+    def all(_object, args, context)
+      scope = Office.all
+
+      scope = scope_with_sort_by(scope, args[:sortBy])
+
+      scope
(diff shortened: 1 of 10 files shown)


liulikun:
Doesn't seem we have a descending sort case. Might be better just sort by ascending by default instead of passing the sort order around.

blakegong:
@liulikun I personally would prefer the `sortBy` approach 😄 The thinking is:

1. Adding one more argument to `offices` query is still very cheap
1. Normally for any resource, people would expect things to be returned in the order of ID, and I'm keen to keep that
1. We might have an advanced office management page in the future, where sorting by any attribute might be needed. I felt like keeping this `sortBy` parameter should be the best option going forward

blakegong:
Thanks for the suggestion @liulikun! Please give this PR a go again 🙇

liulikun:
👍

blakegong:
@liulikun if all good, could you please drop an approval to make this PR mergeable? Or let me know what else is needed. Thank you! 🙇

liulikun (approved):

Issue #67, opened 2018-09-05
Prompt every user to set default office via an IPM


jpandl19:
CSR is having a lot of issues with users having the wrong default office setup. Most people are categorized under "Remote" (since this is what every new user defaults to initially). This means that CSR has to go through each month and manually change the users to their correct office.

We need a one-time in-product message to prompt all current `volunteer_portal` users to update their default office.

Related issue: #66

samzx:
closed by #293

Issue #66, opened 2018-09-05
Prompt users to set default office upon initial login


jpandl19:
CSR is having a lot of issues with users having the wrong default office setup. Most people are categorized under "Remote" (since this is what every new user defaults to initially). This means that CSR has to go through each month and manually change the users to their correct office.

We'd like to prompt the user to set their office immediately when they log in for the first time.

ThuTrinh:
duplicate of https://github.com/zendesk/volunteer_portal/issues/67

PR #109, opened 2018-11-19
Display office name under name of user

blakegong:
## Description

Seems from #66, #67, letting users know of their current office more promptly is an issue. We worked with our designer @shiyunl to come up with some small tweaks to make that more obvious.

## Screenshots

1. Normal
    screenshot 2018-11-19 at 4 29 47 pm

1. With long office name
    screenshot 2018-11-19 at 4 30 03 pm

1. With long user name
    screenshot 2018-11-19 at 4 30 29 pm

1. Normal user mode
    screenshot 2018-11-19 at 4 31 30 pm

## CCs

@rshokrizadeh @dineshsaravanan 

## Risks
Low, pure UI change.

### DIFF
--- a/app/javascript/components/Header/index.js
+++ b/app/javascript/components/Header/index.js
@@ -57,7 +57,10 @@ const Header = ({ currentUser, offices, togglePopover, popover, handleOfficeSele
           <div>
             <button className={s.btn} onClick={togglePopover}>
               <img className={s.img} src={currentUser.photo} />
-              <span className={s.userName}>{currentUser.name}</span>
+              <div className={s.userInfoBox}>
+                <div className={s.userName}>{currentUser.name}</div>
+                <div className={s.userOffice}>{currentUser.office && currentUser.office.name}</div>
+              </div>
               <Popover
                 className={s.popover}
                 open={present(popover)}
--- a/app/javascript/components/Header/main.css
+++ b/app/javascript/components/Header/main.css
@@ -38,25 +38,31 @@
 }
 
 .btn {
-  height: 40px;
-  line-height: 40px;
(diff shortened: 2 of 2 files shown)


dineshsaravanan (approved):

craig-day:
I don't like that it isn't an obvious button. Why would I randomly click on that text? Is there a good reason we can't leave the border?

liulikun (approved):

shiyunl:
Hi @craig-day!

We could certainly leave the border. To add more information to the user selector, we would have the following options: 

displayofficename

The height of the profile dropdown clearly has to increase. If we are ok with having the borders mark out the inconsistent height, we could add the border back…? I've explored slightly different variations of these. 

Alternatively, I have tried adding a dropdown indicator:
with dropdown

But… I'm not too fond of the awkward alignment.

I also dug around a few services and found examples with and without borders. Google applies this somewhat inconsistently, whereas Zendesk goes for the borderless approach.

others

I hear your concern. It is valid. :/ What do you think? Actually, I think we need to have more consistent styling across the board for buttons / drop-downs and such. Accessibility is definitely a concern. If we take some of the lessons from Garden, perhaps we should also consider a hover state…

with hover state

rshokrizadeh:
Got two +1, free to merge!

roseleaf:
fwiw @jianw3i has been our consulting designer on volunteer stuff. may be worth having his 2-cents on the border question too

jianw3i:
In Garden, the icon button doesn't have a border. This is similar to Support top bar icons. https://garden.zendesk.com/react-components/buttons/ 

What @shiyunl designed is 👍

Issue #79, opened 2018-09-20
user search in reporting section


azaiszen:
when you search for a name here, will only search from first name, cant search for surname

not sure if any of the other searches in the portal work like this?

![](https://cl.ly/14b2c31d2732/Screen%20Recording%202018-09-20%20at%2006.02%20pm.gif)

samzx:
closed via #118

PR #118, opened 2018-12-19
 [#79] Make admin tables have a consistent filter

danielbreves:
## Description
This adds a filter to admin tables that didn't have one and makes the filter search any part of the string, not just `startsWith`. It also deletes some unused utils code.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/79)

## CCs
@zendesk/volunteer

## Risks (if any)
* [low] - search may display too many results


xuebingli (approved):
👏

PR #127, opened 2019-01-17
Create a zendesk ticket when individual event is created

### LINKED ISSUE #125: As an Admin, I should get a notification that there is a new 'individual event' that needs to be approved. Ideally a ticket would be created in volunteerhelp.zendesk.com and there would be an approve/decline button in the ticket window
Progress: https://github.com/zendesk/volunteer_portal/compare/sxie/%23125-create-zendesk-ticket

samzx:
Only occurs if the following environment variables are set: 

- `ZENDESK_HOSTNAME` eg. `https://volunteerhelp.zendesk.com`
- `ZENDESK_USERNAME` eg. `admin@zendesk.com`
- `ZENDESK_PASSWORD` eg. `123456`

resolves #125 

## Description
Creates a zendesk ticket so admins of the volunteer portal will get notified of when to approve volunteer events (and possibly assign/delegate to other zendesk agents to verify).

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/125)

## Tests
- [x] unit tests (todo)

Can be done on a zendesk trail account, or zdi (if you dare 🤭)

## Screenshots (if needed)
<details>
<summary>Create individual event</summary>

</details>

<details>
<summary>Ticket automatically created</summary>

</details>

## CCs
@zendesk/volunteer

Anyone specific?

## Risks (if any)
* low - affects creating individual events


app/jobs/zendesk_job.rb
```diff
@@ -0,0 +1,35 @@
...
+    @@username = ENV['ZENDESK_USERNAME']
+    @@password = ENV['ZENDESK_PASSWORD']
+  end
+
+  def self.has_zendesk_integration()
+    return @@hostname && @@username && @@password
+  end
+
+  def self.create_ticket(description)
+    # TODO: async this action, currently takes almost 2s to finish request. Modal is open for too long, confusing users whether action occurrec.
```
xuebingli: Agreed. I didn't think it'd take this long. Sounds like we may need to start using [Active Job](https://guides.rubyonrails.org/active_job_basics.html).

samzx: Maybe a future feature?

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
+require_relative '../test_helper'
+
+SingleCov.covered!
+
+describe 'Zendesk Job' do
+  hostname = 'https://hostname.com'
+  username = 'username@email.com'
+  password = '123456'
```
craig-day: prefer `let` syntax in test for setting up test variables

```rb
let(:hostname) { }
# ...
```

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+  describe 'boostraps zendesk job' do
+    it 'bootstraps by getting environment variables' do
+      ENV['ZENDESK_HOSTNAME'] = hostname
+      ENV['ZENDESK_USERNAME'] = username
+      ENV['ZENDESK_PASSWORD'] = password
```
craig-day: prefer to do test setup in a `before` block. the example (`it`) block should ideally only ever contain assertions

craig-day: this will leave these variables set for the rest of the test suite, leading to possible race conditions if you add a negative test asserting nothing happens when they aren't set.

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      ZendeskJob.bootstrap()
+      assert ZendeskJob.hostname == hostname
+      assert ZendeskJob.username == username
+      assert ZendeskJob.password == password
```
craig-day: use `assert_equal expected, actual` here, otherwise the error message will just say "Assertion failed, no message given." which isn't very helpful.

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+  
+  describe 'has zendesk integration' do
+    it 'should return true if zendesk variables are set' do
+      ZendeskJob.hostname = hostname
+      ZendeskJob.username = username
+      ZendeskJob.password = password
```
craig-day: `before` block

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      assert ZendeskJob.has_zendesk_integration?() == true
```
craig-day: no parens with no args

craig-day: booleans are a good case for using plain `assert`, but adding the second argument as a failure message

```rb
assert ZendeskJob.has_zendesk_integration?, 'must return true'
```

this way when you put together you test context, you get a nicer message like `has zendesk integration :: is true if zendesk variables are set :: must return true`

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      ENV['ZENDESK_PASSWORD'] = password
+      ZendeskJob.bootstrap()
+      assert ZendeskJob.hostname == hostname
+      assert ZendeskJob.username == username
+      assert ZendeskJob.password == password
+    end
+  end
+  
+  describe 'has zendesk integration' do
+    it 'should return true if zendesk variables are set' do
```
craig-day: avoid `should` and `return`. `should` is a reference to an old testing style and is implied, and everything will return.

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  describe 'create ticket' do
+    it 'fails to create ticket if there is no zendesk integration' do
+      ZendeskJob.hostname = nil
+      ZendeskJob.username = nil
+      ZendeskJob.password = nil
+      stub_request(:post, "https://hostname.com/api/v2/tickets.json").
+      with(body: "{\"ticket\":{\"comment\":{\"body\":\"Please visit the volunteer portal as an admin to verify this request.\"},\"subject\":\"An individual event \\\"\\\" has been requested for approval.\"}}",
```
craig-day: just make this a ruby hash and convert it to json instead of manually writing json

test/jobs/zendesk_job_test.rb
```diff
@@ -0,0 +1,61 @@
...
+      assert ZendeskJob.has_zendesk_integration?() == false
+    end
+  end
+
+  describe 'create ticket' do
+    it 'fails to create ticket if there is no zendesk integration' do
+      ZendeskJob.hostname = nil
+      ZendeskJob.username = nil
+      ZendeskJob.password = nil
+      stub_request(:post, "https://hostname.com/api/v2/tickets.json").
```
craig-day: stubbing requests are the perfect example of `before` block setup

craig-day (changes requested):
Thanks for the submission! Quite a few suggestions, let me know if you want me to elaborate on any of the comments.

samzx:
WIP pushed, PR might be stalled in favour of potential "auto-approve" process.

samzx:
#160 aims to solve #125 better in a elegant way. Zendesk Integration could be added later in the future. Will revisit in the future.

Issue #124, opened 2019-01-03
Bypassing admin authorization on client side


xuebingli:
Malicious user could bypass admin authorization by commenting out three lines of client side JS code. 

https://github.com/zendesk/volunteer_portal/blob/f543f99458558115e85abe927a69d8c6eeaa7c5e/app/javascript/pages/admin/Admin/index.js#L92-L94

Once in, they can perform admin actions from the GUI, e.g. approving individual events.

We should enforce server side authorization for admin-only actions.

PR #128, opened 2019-01-17
upgrade graphql-ruby

craig-day:
## Description

This upgrades the graphql-ruby library we use, and then refactors everything to use the Class-based API. This keeps us up to date with the newest library and provides native support for a few nice features such as resolvers, query instrumentation, and authorization. I will be opening up a followup PR to add mutation authorization to address #124. 

## Implementation notes

The upgrade process and rational is documented very well by the author: http://graphql-ruby.org/schema/class_based_api

This is a quick summary of what I did:

- move everything to `app/graphql` from `app/graph`
- use classes an inheritance instead of the `.define` pattern
- mark every field with `null:` instead of the `!` DSL
- mark every argument with `required:` instead of the `!` DSL
- change all `resolve` lambdas to instance methods on the object
- convert all ruby code to use `snake_case`, which the library automatically converts to/from `camelCase` for the client
- replace the graphql controller with the one generated for us

## CCs

@zendesk/volunteer

## Risks (if any)
* High. This includes significant API changes so any given query or mutation could fail.


Gemfile.lock
```diff
@@ -214,6 +214,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-darwin-17
```
thebritican: why is this necessary?

app/controllers/graphql_controller.rb
```diff
@@ -1,20 +1,45 @@
...
+  def execute
+    variables = ensure_hash(params[:variables])
+    context = {
+      current_user: current_user
     }
+
+    render json: PortalSchema.execute(params[:query],
+      variables: variables,
+      context: context,
+      operation_name: params[:operationName])
```
thebritican: why is `:operationName` not snake case at this point?

app/controllers/graphql_controller.rb
```diff
@@ -1,20 +1,45 @@
...
+        ensure_hash(JSON.parse(ambiguous_param))
+      else
+        {}
+      end
+    when Hash, ActionController::Parameters
+      ambiguous_param
+    when nil
+      {}
+    else
+      raise ArgumentError, "Unexpected parameter: #{ambiguous_param}"
```
thebritican: will this eventually return a 4xx or 5xx?

app/graphql/resolvers/event_resolver.rb
```diff
@@ -4,62 +4,60 @@ module EventResolver
...
-      event.title = attrs['title']
-      event.description = attrs['description']
-      event.event_type_id = attrs['eventType']['id']
-      event.organization_id = attrs['organization']['id']
-      event.office_id = attrs['office']['id']
-      event.location = attrs['location']
-      event.capacity = attrs['capacity']
-      event.starts_at = Time.parse(attrs['startsAt'])
-      event.ends_at = Time.parse(attrs['endsAt'])
+    def update_fields(event, input)
```
thebritican: nice rename

app/graphql/types/query_type.rb
```diff
@@ -0,0 +1,97 @@
+module Types
+  class QueryType < Types::BaseObject
+    graphql_name 'Query'
+    description 'The query root for this schema'
+
+    field :node, field: GraphQL::Relay::Node.field, null: true
+    # field :nodes, field: GraphQL::Relay::Node.plural_field, null: true
```
thebritican: delete?

test/graphql/resolvers/event_resolver_test.rb
```diff
@@ -10,19 +10,19 @@
...
-        'organization' => { 'id' => org.id },
-        'office' => { 'id' => office.id },
-        'location' => 'newl',
-        'startsAt' => timestamp.iso8601,
-        'endsAt' => (timestamp + 10.minutes).iso8601,
-        'capacity' => 10,
-      }
-
-      EventResolver.create(nil, {input: attrs}, nil)
+      input = stub(
```
thebritican: why are we stubbing now?

gregziegan (approved):
Nice work! I didn't see any typos or issues with the approach. Schema looks well documented and consistent. Some small questions and nits but otherwise ship it after some manual testing that things load. Do we have GraphQL API tests at all? Or a schema linter? (have a script catch if we're broken the schema)

Gemfile.lock
craig-day: This is just a newer bundler config thing. It isn't "necessary" but its something that gets automatically added with newer bundler versions.

app/controllers/graphql_controller.rb
craig-day: I believe this is before the query has been processed and the conversion has happened. We are in a controller here, not yet in the graphql resolution portion. This controller was auto generated by the new gem.

app/controllers/graphql_controller.rb
craig-day: This will return a 500. This was the default behavior generated by the gem, but we will probably want to add consistent and explicit error handling for graphql errors. #130

app/graphql/types/query_type.rb
craig-day: This was carried over from the previous type definition. #131

app/controllers/graphql_controller.rb
thebritican: i thought so but didn't know if we could move the camelCase -> snake_case step earlier via a library function

app/controllers/graphql_controller.rb
thebritican: thanks for tracking

test/graphql/resolvers/event_resolver_test.rb
thebritican: forgive my forgetfulness about `stub`. Is it any different from using `OpenStruct.new()`? Any performance penalties? If we're not using the `stub` API methods, maybe tests are faster with a simple struct?

craig-day: `stub` is a common testing pattern. in other frameworks it may be called `mock` or `double`. I don't want to switch to OpenStruct because it's not very common. Also OpenStruct is different that Struct and I don't think it would be any faster than a stub.

app/controllers/graphql_controller.rb
craig-day: I'm not aware of them exposing a nice function for it. I mean, we have activesupport so we could call `underscore` if we wanted to, but that would involve manipulating the hash. I'm just not sure if its worth manipulating the data before forwarding it into the graphql execution layer.

PR #133, opened 2019-01-23
require an admin to C_UD most resources

### LINKED ISSUE #124: Bypassing admin authorization on client side

craig-day:
## Description
This does two primary things. First, it converts every mutation to a class to map nicely to each GraphQL object that it is backing. Second, it adds a little helper to take advantage of the `Mutation#ready?` hook to perform authorization on most of the mutations. This will restrict the Create, Update, and Delete actions to admins for every resource with a few exceptions: 

- individual events are scoped to and managed by the current user
- the default office for a user is managed by that user
- signups can be created and destroyed by any user, but are scoped to the current user

resolves #124

it event has a nice little error message and valid JSON 

screen shot 2019-01-23 at 3 05 01 pm

## CCs

@zendesk/volunteer

## Risks
* High - mutations could fail. This means it may break when trying to create/update/delete almost anything from the admin pages. Signing up for an event should be untouched.

### DIFF
--- a/app/graphql/mutations/base_mutation.rb
+++ b/app/graphql/mutations/base_mutation.rb
@@ -1,4 +1,4 @@
 module Mutations
-  class BaseMutation < GraphQL::Schema::RelayClassicMutation
+  class BaseMutation < GraphQL::Schema::Mutation
   end
 end
--- a/app/graphql/mutations/base_mutation.rb
+++ b/app/graphql/mutations/base_mutation.rb
@@ -1,4 +1,8 @@
 module Mutations
   class BaseMutation < GraphQL::Schema::Mutation
+
+    def self.require_admin
+      self.include Concerns::RestrictToAdmin
+    end
   end
 end
--- a/app/graphql/mutations/create_edit_individual_event.rb
+++ b/app/graphql/mutations/create_edit_individual_event.rb
@@ -1,0 +1,28 @@
+module Mutations
+  class CreateEditIndividualEvent < BaseMutation
+    description "Create a new individual event for a logged-in user"
+
+    null true
+
+    argument :input, Types::Input::CreateEditIndividualEventInputType, required: true
+
+    def resolve(input:)
+      if input.id
+        IndividualEvent.find(input.id)
+      else
+        IndividualEvent.new
+      end.tap do |individual_event|
+        individual_event.description = input.description
+        individual_event.office_id = input.office_id
+        individual_event.user = context[:current_user]
+        individual_event.date = Time.at(input.date)
+        individual_event.duration = input.duration
+        individual_event.event_type_id = input.event_type_id
+        individual_event.organization_id = input.organization_id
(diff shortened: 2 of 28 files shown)


app/graphql/mutations/base_mutation.rb
```diff
@@ -1,4 +1,8 @@
 module Mutations
-  class BaseMutation < GraphQL::Schema::RelayClassicMutation
+  class BaseMutation < GraphQL::Schema::Mutation
```
thebritican: what's the **TL;DR** difference between `RelayClassMutation` and `Mutation`?

app/graphql/mutations/create_edit_individual_event.rb
```diff
@@ -0,0 +1,28 @@
+module Mutations
+  class CreateEditIndividualEvent < BaseMutation
+    description "Create a new individual event for a logged-in user"
+
+    null true
+
+    argument :input, Types::Input::CreateEditIndividualEventInputType, required: true
+
+    def resolve(input:)
```
thebritican: test? what happens if `IndividualEvent.find()` fails?

app/graphql/mutations/reject_individual_event.rb
```diff
@@ -0,0 +1,18 @@
...
+    require_admin
+
+    description 'Mark an individual event as rejected'
+
+    null true
+
+    argument :id, ID, required: true
+
+    def resolve(id:)
+      event = IndividualEvent.find(id)
```
thebritican: same as above throughout PR

gregziegan (commented):
looks good. curious about the `nil` case across the resolvers

app/graphql/mutations/create_edit_individual_event.rb
xuebingli: I believe it's going to lead to a 500 server side error.

app/graphql/mutations/base_mutation.rb
craig-day: `RelayClassMutation` expects a certain form of return type, but we don't follow that form because we don't follow relay standards currently.

app/graphql/mutations/create_edit_individual_event.rb
craig-day: I added a case for this so we now get a nice error

```
{
  "errors": [
    {
      "message": "Couldn't find IndividualEvent with 'id'=999"
    }
  ]
}
```

gregziegan (approved):
nice work!

PR #140, opened 2019-03-09
Adds CircleCI test output analysis

screamish:
Adds CircleCI test output via the minitest-ci gem for ruby tests and jest-junit for javascript tests.

## Description
I noticed that this project doesn't have test output integration with CircleCI yet. This is my first ever PR to this project so I was looking for something small and peripheral. I would have marked this as a DRAFT PR since it's definitely not ready to be merged, but I need it to get picked up by CircleCI.

## References
https://circleci.com/docs/2.0/collect-test-data/

I'm very rusty on Ruby/Rails so please forgive me if I'm just missing some obvious things here.

## Tasks
- [x] Resolve conflict between Maxitest and Minitest plugin loading

## CCs
@zendesk/volunteer @craig-day @grosser 

## Risks
* low - only adjusts CI test reporting, no change to application


screamish:
When maxitest/autorun is removed from the test_helper.rb I get the following:

```
❯ TESTOPTS="-h" bundle exec rake test
minitest options:
    -h, --help                       Display this help.
    -s, --seed SEED                  Sets random seed. Also via env. Eg: SEED=n rake
    -v, --verbose                    Verbose. Show progress processing files.
    -n, --name PATTERN               Filter run on /regexp/ or string.
    -e, --exclude PATTERN            Exclude /regexp/ or string from run.

Known extensions: ci, rails, pride
        --ci-report                  Enable Minitest::Ci Reporting.
```

With require maxitest/autorun back in test_helper (and no other changes at all), I get the following:

```
❯ TESTOPTS="-h" bundle exec rake test
minitest options:
    -h, --help                       Display this help.
    -s, --seed SEED                  Sets random seed. Also via env. Eg: SEED=n rake
    -v, --verbose                    Verbose. Show progress processing files.
    -n, --name PATTERN               Filter run on /regexp/ or string.
    -e, --exclude PATTERN            Exclude /regexp/ or string from run.

Known extensions: maxitest_verbose_backtrace
```

grosser:
`bundle update maxitest` to make sure you got latest

grosser:
what is this test output used for ?

screamish:
Thanks @grosser, updating maxitest seems to have fixed the plugin loading!

I believe this page https://circleci.com/blog/how-to-output-junit-tests-through-circleci-2-0-for-expanded-insights/ is the best quick overview of what providing test output to circleci in this format offers. I'm certainly not trying to suggest it's a must-have for any project, though, and won't be offended at all if this PR gets closed and not merged.

.gitignore
```diff
@@ -24,3 +24,6 @@
 vendor/gems
 /public/packs
 /node_modules
+
+# Ignore test reports
+/test/reports/
```
grosser: if it used tmp/test_reports, then they would not be in the way when searching for code in the test directory

grosser (approved):
looks good / safe
if possible see if there is a way to make package lock not be such a giant diff, maybe there is a `update --conservative` option like bundler has

.gitignore
screamish: Good idea, done.

screamish:
> if possible see if there is a way to make package lock not be such a giant diff, maybe there is a `update --conservative` option like bundler has

I'm no npm expert, but AFAIK the extent to which dependencies are "pinned" or "floating" during operations like adding a new dependency and installing it are controlled by the [way dependency versions are specified in the package.json](https://docs.npmjs.com/files/package.json#dependencies).  All our deps are listed with the caret `^` prefix which means [all but the left-most non-zero version number will float](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004).

If anyone knows otherwise, please correct me.

xuebingli (approved):
Thanks for the PR! Looking forward to the insights! 👏

Issue #110, opened 2018-11-19
Small design tweaks


shiyunl:
### Alignment for Dashboard table is off
![designbug-tab navigation styling](https://user-images.githubusercontent.com/7077683/48695843-327b1780-ec1b-11e8-9189-5c738c92bcce.png)

### Navigation tab styling 
![designbug-alignment in volunteer table](https://user-images.githubusercontent.com/7077683/48695857-34dd7180-ec1b-11e8-86be-414ab802fd77.png)

PR #142, opened 2019-03-13
Fixed visual bugs as outlined in issue #110 

samzx:
Fixed visual bugs as outlined in issue #110 

## Description
* Fixed the border bottom on the tab menus
* Fixed style alignment bug for named avatars and hours on leaderboards

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/110)

The user.group field is unused (UserEntry graphQL query does not retrieve this field).
The NamedAvatar component `subtitle` field is therefore unused.

## Screenshots (if needed)
<details>
<summary>Before</summary>

image

image

</details>

<details>
<summary>After</summary>

image

image

</details>

## CCs

@zendesk/volunteer 

## Risks (if any)
* low

### DIFF
--- a/app/javascript/components/UserList/index.js
+++ b/app/javascript/components/UserList/index.js
@@ -14,6 +14,7 @@ const UserList = ({ users, destroySignup }) => (
             userId={user.id}
             image={user.photo}
             name={user.name}
+            subtitle={user.group}
             showRemove={present(destroySignup)}
             onRemove={() => destroySignup(user)}
           />
--- a/app/javascript/components/UserList/index.js
+++ b/app/javascript/components/UserList/index.js
@@ -14,7 +14,6 @@ const UserList = ({ users, destroySignup }) => (
             userId={user.id}
             image={user.photo}
             name={user.name}
-            subtitle={user.group}
             showRemove={present(destroySignup)}
             onRemove={() => destroySignup(user)}
           />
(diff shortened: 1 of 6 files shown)


xuebingli:
It'd be great if you could attach a screenshot.

app/javascript/components/UserList/index.js
```diff
@@ -14,7 +14,6 @@ const UserList = ({ users, destroySignup }) => (
             userId={user.id}
             image={user.photo}
             name={user.name}
-            subtitle={user.group}
```
xuebingli: I did some digging and it turns out user does have a field called `group`, though not all users have that field populated.

```ruby
irb(main):007:0> User.where(group: nil).count
=> 1206
irb(main):008:0> User.where.not(group: nil).count
=> 1712
```

samzx: That's interesting. What is an example of a `group`?

We could either add the meta data to the subtitle section by fixing the graphQL `UserEntry` query or leave it as is (removing the subtitle section, as it's empty anyway).

It may as well be too much clutter if the group information isn't interesting or relevant anyway.

Thoughts?

xuebingli: > What is an example of a group?

"Growth Marketing"

> We could either add the meta data to the subtitle section by fixing the graphQL UserEntry query or leave it as is (removing the subtitle section, as it's empty anyway).

I reckon we should fix the GraphQL query and populate `group`, since more than half of users have `group` field populated.

samzx: @xuebingli fixed!

xuebingli (approved):

PR #148, opened 2019-03-27
Added clone event functionality

### LINKED ISSUE #16: Add "duplicate event" feature to streamline event creation
Currently, we have a table for storing some reoccurring events in the "event_groups" table. 
The events table also contains a column "event_groups_id" which we can use to relate the two. 

We need a way for an admin to mark an event as "reoccurring", creating a record in the events_group_table, with the title as the title of the event. Then update that new event_group_id in that event's record.

samzx:
## Short description

closes #16 
- New event accepts url parameters (event-id) for pre-fill
- Event dashboard ability to clone event by linking to above mentioned url `.../portal/admin/events/new/78` would pre-fill 'new event' page with the contents of event id 64.

## Description
Admins would like to create clone events that occur frequently without having to fill in the same information again. In particular, reoccuring events. This will let admins clone existing events by:
- Going to the Event admin page
- Clicking clone next to the event
- They will be prompted to create a new event, with the data pre-filled (cloned)

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/16)

- Modifies the new event page to accept a url parameter to pre-fill data
- Optional url parameter field, if no parameter is provided, it is the normal new event page

## Screenshots (if needed)

image

☝️click clone
👇pre-fills data (located at `http://localhost:5000/portal/admin/events/new/78`)

image

## CCs

@zendesk/volunteer

## Risks (if any)
* low

### DIFF
--- a/app/javascript/pages/admin/Events/index.js
+++ b/app/javascript/pages/admin/Events/index.js
@@ -24,6 +24,7 @@ const eventsSort = 'STARTS_AT_DESC'
 const actionLinks = (event, deleteEvent) => (
   <div className={s.actionColumn}>
     <Link to={`/portal/admin/events/${event.id}/edit`}>Edit</Link>
+    <Link to={`/portal/admin/events/new/${event.id}`}>Clone</Link>
     <button className={`${s.deleteAction}`} onClick={() => deleteEvent(event)}>
       Delete
     </button>
@@ -66,7 +67,7 @@ const columns = deleteEvent => [
     Header: 'Actions',
     accessor: 'id',
     sortable: false,
-    width: 130,
+    width: 150,
     Cell: ({ original }) => actionLinks(original, deleteEvent),
   },
 ]
@@ -200,11 +201,8 @@ function mapStateToProps(state, _ownProps) {
   }
 }
 
-const withActions = connect(
-  mapStateToProps,
-  {
-    graphQLError,
-  }
-)
+const withActions = connect(mapStateToProps, {
+  graphQLError,
+})
 
 export default withActions(withData(Events))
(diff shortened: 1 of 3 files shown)


app/javascript/pages/admin/Events/index.js
```diff
@@ -21,16 +21,17 @@ import 'style-loader!css-loader!react-table/react-table.css'
 
 const eventsSort = 'STARTS_AT_DESC'
 
-const actionLinks = (event, deleteEvent) => (
+const actionLinks = (event, deleteEvent, cloneEvent) => (
```
xuebingli: `cloneEvent` doesn't appear to be used. I reckon we could remove it.

samzx: thanks, I forgot to clean this one up! 👍

app/javascript/pages/admin/Events/new.js
```diff
@@ -15,11 +15,32 @@ import CreateEventMutation from './mutations/create.gql'
 
 import Loading from 'components/LoadingIcon'
 
-const NewEvent = ({ createEvent, data: { networkStatus, eventTypes, offices, organizations } }) =>
+const transformToReduxFormState = event => {
```
samzx: A note with whether to use graphQL to get the below shape for the redux-form state:
- I don't think it's worth it, as you'd need to create another graphQL query (duplicate query, and extra latency) to get the same information the existing query captures (existing query gets all the necessary information, and more - for form options)
- You'll still need to create a function to parse the dates retrieved from graphQL to a javascript `Date` object in which the redux-form state is configured.

This solution uses the existing event information retrieved and transforms it to the shape that redux-form is configured with.

xuebingli (commented):
👏

xuebingli (approved):

PR #160, opened 2019-05-22
Admins can bulk approve or reject individual events.

### LINKED ISSUE #155: As an admin, I want a faster approval process.
- Some suggestions include auto-approve toggle in settings. (no admin intervention)
- Select all -> approve action
- Plain approve all button

samzx:
resolves #155

## Description

Currently admins, have to click "approve" or "reject" on each individual event to complete their action. This gets tedious and time consuming when admins have to go through a large list (especially if network is slow).

It is noted from admins that they typically approve all outcomes and only occasionally don't approve when users enter a wrong value - for example, they entered hours instead of minutes.

It was suggested to have an auto-approval process: #125 This PR is a solution in between, where approval does require some oversight (such as the before mentioned example) to catch mistakes, but speeds up the approval process immensely.

Also converted tables to React tables for sorting and adjustable column widths.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/155)

## Implementation notes
The bulk of it is renaming to plural, and converting to react tables

## Screenshots
<details>
<summary>Before</summary>

image

</details>

### After

bulkactiontrim

## CCs
@zendesk/volunteer

## Risks (if any)
* low


app/javascript/pages/admin/IndividualEvents/index.js
```diff
@@ -3,82 +3,168 @@ import { compose, graphql } from 'react-apollo'
...
+  render() {
+    const { data } = this.props
+    const { networkStatus, pendingIndividualEvents } = data
+    return networkStatus === NetworkStatus.loading ? (
+      <Loading />
+    ) : (
+      <div className={s.eventsTable}>
+        <h3>Pending Approval</h3>
+        <div>
+          <select
```
thebritican: could use Garden's select here?

app/javascript/pages/admin/IndividualEvents/main.css
```diff
@@ -178,3 +178,20 @@ textarea {
 .muiTextField hr {
   display: none;
 }
+
+.eventsTable > div {
```
thebritican: instead of coupling to a particular structure & element, why not make a class specific for that `div` and give it a name?

app/javascript/pages/admin/IndividualEvents/main.css
```diff
@@ -178,3 +178,20 @@ textarea {
...
+  border: none;
+}
+
+ .eventsTable > div > div:nth-child(1) > div:nth-child(1) {
```
thebritican: same as above comment. I'm not sure what `eventsTable>div[1][1]` means at a glance

gregziegan (approved):
just some nitpicks. Nice work!

app/javascript/pages/admin/IndividualEvents/index.js
samzx: I don't think we have garden added to this project yet, it's a good idea though. I'd be happy to do a PR later on to redo components to garden components. All the UI isn't too consistent either, forms etc. So might bundle those changes later on.

It's an admin interface so I feel it should be fine for now.

app/javascript/pages/admin/IndividualEvents/main.css
samzx: Same with the comment below, but it's because by using react tables and I wasn't able to tag elements under the api. But after a second look, seems like you can pass `ReactTable` a `style` prop which would be much cleaner. I'll see if I can do that instead.

samzx: Okay, had another look. It seems like the api only allows you to specify styles only for a columns. I think the reason I first did that was because of the css-loader made it hard to target the actual class name ie. 'ReactTable'. I figured it out though 🎉

PR #162, opened 2019-05-29
Add SAML Authentication

adam-hart:
## Description

This is a re-work of https://github.com/zendesk/volunteer_portal/pull/150.  I realized that the previous PR wasn't _actually_ SAML, but just OAuth mediated by Okta.  This introduces an actual client-initiated SAML authentication solution.  Like https://github.com/zendesk/volunteer_portal/pull/150, it will default to Google OAuth if the secrets are available, otherwise we will use SAML.

SAML metadata will be stored in the `User` model under a `metadata` column.  [It's a Postgres `jsonb` type](https://www.postgresql.org/docs/9.5/functions-json.html), which has well supported query features.  Its integration into reporting will come later.

@zendesk/volunteer 

## Risks
* Medium:  Could break Google authentication


app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -4,18 +4,23 @@ class OmniauthCallbacksController < ActionController::Base
...
-      return
-    end
+    set_user_photo
+    set_user_metadata
+
+    user.save!
+
+    session[:user_id] = user.id
+
+    redirect_to params[:return_to] || :portal
```
craig-day: we should do some validations on the return_to param. At least make sure it's only a relative URL otherwise this can be used as an attack vector for someone to get people onto a malicious website.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -30,39 +35,42 @@ def logout
...
+  def user_signed_in?
+    session[:user_id].present? && User.find_by(id: session[:user_id])
+  end
+  helper_method :user_signed_in?
 
-    redirect_to params[:return_to] || :portal
-  rescue ActiveRecord::RecordInvalid => e
-    render text: e.message, status: :forbidden
+  def verify_callback_contents
+    raise 'Authentication provider must provide email, first_name, and last_name parameters' unless (auth_info.keys & ['email', 'first_name', 'last_name']).size == 3
```
craig-day: line is pretty long, maybe turn into an early return guard or multiline the unless?

app/views/omniauth_callbacks/login.html.erb
```diff
@@ -9,6 +9,10 @@
   </div>
 
   <div class="login-form">
-    <%= button_to 'Sign in with Google', '/auth/google_oauth2', class: 'google-signin-btn' %>
+    <% if ENV['GOOGLE_CLIENT_ID'] %>
+      <%= button_to 'Sign in with Google', '/auth/google_oauth2', class: 'signin-btn google' %>
+    <% else %>
+      <%= button_to 'Sign in with SAML', '/auth/saml', class: 'signin-btn okta' %>
+    <% end %>
```
craig-day: Do we want to if/else here? Or should we just allow both if the settings are there?

config/puma.rb
```diff
@@ -2,6 +2,8 @@
 threads_count = Integer(ENV['RAILS_MAX_THREADS'] || 5)
 threads threads_count, threads_count
 
+worker_timeout 600
```
craig-day: :-1: this should at a minimum be configurable if we want to add it at all

docs/deploy.md
```diff
@@ -26,26 +26,44 @@ You'll need them in the steps to come. The portal's name should be concise and d
...
 
 2. Use a custom domain name.
 
-You can connect any custom domain name you own to your Heroku deployment. We'll go over how to link the two later in the setup, but make sure you know what it is!
+  You can connect any custom domain name you own to your Heroku deployment. We'll go over how to link the two later in the setup, but make sure you know what it is!
 
-### Set up Google OAuth
+### Set up User Authorization
+
+Volunteer Portal currently supports two forms of user authentication:  Google Oauth and SAML.  Choose whichever one makes most sense for you, and follow the instructions below to get started.  If you have requests for more authentication methods, feel free to let us know by joining our Slack channel ([the link is in our main README](https://github.com/zendesk/volunteer_portal/blob/ahart/okta-auth/README.md#volunteer-portal)), or follow [our developer documentation](https://github.com/zendesk/volunteer_portal/blob/ahart/okta-auth/docs/development.md) to add it yourself!
```
craig-day: probably don't want to link to this branch, just use relative URLs for these links.

craig-day (changes requested):
Looks good. I don't see why we can't or shouldn't support both simultaneously. I think that should just involve changing the `if/else` statements to multiple `if` checks to add each thing that is configured.

config/puma.rb
adam-hart: Ah good catch; I did this for debugging and forgot to remove it.

config/initializers/omniauth.rb
```diff
@@ -1,64 +1,41 @@
...
-  # If you need to acquire whether user picture is a default one or uploaded by user.
-  #
-  # provider :google_oauth2, ENV['GOOGLE_KEY'], ENV['GOOGLE_SECRET'], {
-  #   :skip_image_info => false
-  # }
+  if ENV['SAML_ISSUER']
+    provider :saml,
+    :issuer                             => ENV['SAML_ISSUER'],
+    :idp_sso_target_url                 => ENV['SAML_IDP_SSO_TARGET_URL'],
+    :idp_cert                           => ENV['SAML_IDP_CERT']
```
craig-day: these lines should be indented, right? and we probably don't need so much space between key and value. and you can just use `key: value` instead of `:key => value`

craig-day: should these be `ENV.fetch` calls so they blow up if the config is not set?

config/initializers/omniauth.rb
```diff
@@ -1,64 +1,41 @@
...
+      redirect_uri: redirect_uri,
+      setup: ->(env) do
+        env['omniauth.strategy'].options['token_params'] = {
+          redirect_uri: redirect_uri
+        }
+      end
+    }
+
+    if domain = ENV['OAUTH_DOMAIN']
+      options.merge!(hd: domain)
```
craig-day: single key merge is weird, could be 

```rb
options[:hd] = ENV['OAUTH_DOMAIN'] if ENV['OAUTH_DOMAIN']
```


or if you want to keep the if assignment, just set the key instead of merging

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -30,39 +39,43 @@ def logout
...
-    user.locale       = request.env.dig('omniauth.auth', 'extra', 'raw_info', 'locale')
+  def set_user_photo
+    user.photo = auth_info['image'] if auth_info['image'].present?
+  end
 
-    user.save!
+  def set_user_metadata
+    user.metadata = saml_response_attributes.attributes.keys.each_with_object({}) do |key, metadata|
+      # Ignore default fields we already capture in user or don't care about
+      next if ['first_name', 'last_name', 'email', 'fingerprint'].include?(key)
```
craig-day: should this array be a constant?

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -30,39 +39,43 @@ def logout
...
+  def user_signed_in?
+    session[:user_id].present? && User.find_by(id: session[:user_id])
+  end
+  helper_method :user_signed_in?
 
-    redirect_to params[:return_to] || :portal
-  rescue ActiveRecord::RecordInvalid => e
-    render text: e.message, status: :forbidden
+  def verify_callback_contents
+    return if (auth_info.keys & ['email', 'first_name', 'last_name']).size == 3
```
craig-day: same here? should it be a constant? Maybe we can reuse the constant between this and `set_user_metadata`

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,21 +1,33 @@
 class OmniauthCallbacksController < ActionController::Base
+  REQUIRED_PARAMS = ['email', 'first_name', 'last_name'].freeze
+  METADATA_IGNORED_PARAMS = REQUIRED_PARAMS + ['fingerprint'].freeze
```
craig-day: I _think_ this is just freezing the `fingerprint` array, and not the resulting `METADATA_IGNORED_PARAMS` params. Might want to wrap in parens, or use `foo.concat(bar).freeze`

craig-day (approved):
might want to fix the frozen array thing, but looks good! :+1:

app/controllers/omniauth_callbacks_controller.rb
adam-hart: You know, I thought about that and then thought, "eh it's concatenating two frozen arrays" but I suppose it's still mutable so I should fix it.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -1,21 +1,33 @@
 class OmniauthCallbacksController < ActionController::Base
+  REQUIRED_PARAMS = ['email', 'first_name', 'last_name'].freeze
+  METADATA_IGNORED_PARAMS = (REQUIRED_PARAMS + ['fingerprint']).freeze
```
craig-day: ❤️ thanks

PR #176, opened 2019-07-31
Properly Check If Apollo Client Is Loading

xuebingli:
The current way of checking if Apollo Client is still loading, which relies on network status, is flawed. This PR proposes a fix.

## Description

We currently rely on network status provided by Apollo Client to check if it is still loading. For example, 

```javascript
if (networkStatus === NetworkStatus.loading) { ... }
```

This method is flawed. `NetworkStatus.loading` in Apollo Client has very specific meaning. Apollo Client has five "loading" statuses. Namely, `loading`, `setVariables`, `fetchMore`, `fetchMore`, and `poll`. See https://github.com/apollographql/apollo-client/blob/d8f5cb8d46cb30e49cebea7886ac99472b46ad59/packages/apollo-client/src/core/networkStatus.ts#L5-L10 for details.

The consequence is that sometimes page rendering is triggered when the data is not fully loaded, leading to null pointer errors on the frontend.

This PR proposes a fix that relies on the `loading` indicator, for proper loading status check.

## References
[Rollbar Exception](https://rollbar.com/zd-volunteer-portal/zd-volunteer-portal/items/7/)

## CCs

@zendesk/volunteer

## Risks (if any)
* [medium] - break page rendering due to incorrect loading check result

### DIFF
--- a/app/javascript/pages/App/index.js
+++ b/app/javascript/pages/App/index.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react'
 import R from 'ramda'
 import { compose, graphql } from 'react-apollo'
-import { NetworkStatus } from 'apollo-client'
 import { connect } from 'react-redux'
 import injectTapEventPlugin from 'react-tap-event-plugin'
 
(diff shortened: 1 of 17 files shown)


craig-day:
We don't want `loading` to be true for `fetchMore` or `poll`, which is explicitly why we were only checking the one state in the past. We only want this state when we have never issued a query before and we have no data because we show a loading indicator instead of the calendar. We don't want to replace the calendar with a loading indicator every time we fetch more data or periodically update. I would argue the new approach is more flawed than the old approach.

ThuTrinh (approved):

ThuTrinh:
@samzx , I think we should close this one as it will be superceded by the apollo hooks/garden sweep.

samzx:
Agreed, @ThuTrinh closing in favour of apollo hooks/garden sweep

PR #182, opened 2019-08-14
[WIP]Upgrade React to 16.9.0

RomainEndelin:
Upgrading React to 16.9.0, with associated dependencies

## Description

Upgrading React is useful for 3 things:
- It will facilitate other major packages upgrades (Apollo, etc.)
- It _may_ help us switch to Garden
- We can use hooks, and more easily remove Redux

## Implementation notes

**Upgraded dependencies:**

- `react-google-maps`: `^4.11.0` => `^9.4.5`
- `react-router`: `^2.6.0` => `^3.2.0`
- `react-big-calendar`: `^0.12.3` => `^0.22.0`
- `material-ui`: `^0.16.5` => `^0.20.0`

**Discussions:**

- I added `sourcemaps` in development. It's out-of-scope from this PR, and if you'd like I can open a separate PR for it.
- I removed `React-tap-event-plugin`, since it seems to have become unnecessary, mostly unmaintained, and possibly breaking compatibility
- Upgrading `prop-types` probably wasn't required

**New deprecations:**

Many new deprecations have appeared, regarding React lifecycle. These are caused by out-dated dependencies.
These are harmless for now, but things will break if we try to upgrade React 17 (one day).

I didn't look too much into it. Some of these dependencies are not worth upgrading since we'll replace them (Redux, Material UI, etc.). Others should be upgraded in another PR (React-Apollo, React-Router).

Screenshot 2019-08-15 at 11 55 07

**Breaking changes:**

The UI might have some tiny differences (as highlighted in Jest snapshot tests)

## References
[Github Issue #235](https://github.com/zendesk/volunteer_portal/issues/235)

## CCs

@xuebingli @samzx @ThuTrinh 

## Risks
* medium
  - A major package upgrade may have introduced unnoticed breaking changes
  - Some dependencies between packages may be broken


app/javascript/components/Calendar/index.js
```diff
@@ -158,4 +159,4 @@ const Calendar = ({
     </Layout>
   )
 
-export default Calendar
+export default MyCalendar
```
xuebingli: Could we `import { Calendar as BigCalendar } from 'react-big-calendar'` to avoid renaming the component? Otherwise we introduce inconsistency between component name and file name.

app/javascript/pages/App/index.js
```diff
@@ -14,7 +13,6 @@ import UpdateUserOfficeMutation from './UpdateUserOfficeMutation.gql'
 
 class AppPage extends Component {
   constructor(props) {
```
xuebingli: I reckon we can remove `constructor` altogether, since it's not doing anything other than calling `super(props)`.

xuebingli (approved):
Very well done, Romain! 👏 

Let me test this change in our next session before merge to make sure nothing breaks.

app/javascript/pages/App/index.js
RomainEndelin: Haha yes, good catch!

app/javascript/components/Calendar/index.js
RomainEndelin: Sure, good idea (actually I did this as a temporary workaround, and then forget about it 🙁 )

app/javascript/pages/App/index.js
RomainEndelin: I turned it into a function component as well 🙂

samzx (changes requested):
A couple bugs I ran into:
1. Material UI drop downs, no longer typable. Can't filter the drop down by typing.
image


2. Creating event type (and events, or it seems like most forms don't forward the page) - ends up sending lots of graphql queries.

image

3. Can't create events any more: clicking save doesn't do anything. (doesn't forward to next page or create event) - there's some errors in the console.
image

Just a few I picked up and recorded. There may be more in the admin section that might be worth looking at


Number 2 and 3 might be caused by the same error:
```
 error      Error: Given action "GRAPHQL_MUTATION_ERROR", reducer "graphQLErrors" returned undefined. To ignore an action, you must explicitly return the previous state. If you want this reducer to hold no value, you can return null instead of undefined.
    at combination (combineReducers.js:123)
    at dispatch (createStore.js:165)
    at index.js:104
    at index.js:14
    at Object.graphQLError (bindActionCreators.js:3)
    at new.js:100
```

RomainEndelin:
Thanks for the report @samzx ! It turns out 1. is a pretty big issue: Material UI 0.20's autocomplete is incompatible with React 16!

We don't want to update Material above 0.20 because these are huge breaking changes. So instead, I will take some time and try to replace _AutoComplete_ in our code by Garden components.

sunstryder:
> Thanks for the report @samzx ! It turns out 1. is a pretty big issue: Material UI 0.20's autocomplete is incompatible with React 16!
> 
> We don't want to update Material above 0.20 because these are huge breaking changes. So instead, I will take some time and try to replace _AutoComplete_ in our code by Garden components.

Since we have an issue for upgrading to use Garden, which itself has a dependency of React 16, should we unblock this PR? @samzx @RomainEndelin @xuebingli

RomainEndelin:
Yes @sunstryder, as far as I can tell, we'll have to upgrade React-big-calendar, MaterialUI (to be partially replaced by Garden, at least for dropdowns) and React at the same time:

For React-big-calendar, their documentation isn't great, and I didn't find a version which is compatible with both React 15 & 16.
For Material UI, 0.20 has React 15-16 compatibilty, but it seems to break against auto-complete. It would be a good opportunity to replace these components by Garden; but AFAIK Garden isn't compatible with React 15, so we'll also need React 16.

Also, React-Router upgrade seems to be broken too, I'm looking into it 👀.

To keep this PR as small as possible, I'm trying to upgrade everything I can prior to React 16 in https://github.com/zendesk/volunteer_portal/pull/208

app/javascript/components/EventForm/index.js
```diff
@@ -75,6 +76,22 @@ const renderField = props => {
...
+  // return (
+  //   <GardenField>
+  //     <Label>Autocomplete with debounce</Label>
+  //     <Hint>This example includes basic debounce logic using Hooks</Hint>
+  //     {/* <label className={s.label}>
+  //       {label} <span className={s.errorMsg}>{touched && error ? renderError(error) : '*'}</span>
+  //     </label> */}
+  //     <div>{Custom ? <Custom {...props} /> : fieldInput}</div>
+  //   </GardenField>
+  // )
```
sunstryder: Should we delete this commented out return?

app/javascript/components/EventForm/index.js
```diff
@@ -96,21 +113,85 @@ const dateTimeSplitChange = (part, currentValue, newValue, callback) => {
...
+  //   dataSource={organizations}
+  //   dataSourceConfig={{ text: 'name', value: 'id' }}
+  //   filter={AutoComplete.fuzzyFilter}
+  //   onNewRequest={(chosen, _i) => onChange(chosen.id)}
+  //   className={s.muiTextField}
+  //   textFieldStyle={styles.muiTextField}
+  //   menuStyle={{ overflowY: 'scroll', height: 200 }}
+  //   openOnFocus
+  //   fullWidth
+  // />
```
sunstryder: Is there a reason for leaving in these commented lines?

sunstryder (approved):
Nice :) Looking forward to this going in

kayison:
Is this ok to close @RomainEndelin? These changes should be covered off in #252 and #260 #266.

RomainEndelin:
Sure! Sorry @kayison for not carrying on with this PR, I think it was a bit too ambitious for my react skills... Thanks 🙂

PR #183, opened 2019-08-14
Disable Apply button when no bulk action is selected

### LINKED ISSUE #169: Bulk action 'apply' should not be clickable when an action is not selected
You can click "apply" when no action is selected (it would apply a 'no action' and de-select)

samzx:
## Description
Currently, admins can press the 'Apply' button even when no bulk action is selected. This will 'action' and remove all the selected events, and nothing will happen. To prevent confusion, we should just disable this button.

## References
Resolves https://github.com/zendesk/volunteer_portal/issues/169

## Screenshot

image

## Risks (if any)
Low

### DIFF
--- a/app/javascript/pages/admin/IndividualEvents/index.js
+++ b/app/javascript/pages/admin/IndividualEvents/index.js
@@ -112,6 +112,7 @@ class IndividualEvents extends React.Component {
           </select>
           <button
             className={s.bulkActionButton}
+            disabled={!this.state.bulkAction}
             onClick={() => {
               const { selected, bulkAction } = this.state
               if (selected.length <= 0) return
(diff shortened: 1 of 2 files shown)


app/javascript/pages/admin/IndividualEvents/index.js
```diff
@@ -112,6 +112,7 @@ class IndividualEvents extends React.Component {
           </select>
           <button
             className={s.bulkActionButton}
+            disabled={!this.state.bulkAction}
```
xuebingli: I'm not sure about the use of `state` here. Ideally state management should be centralized in Apollo. That said, state management is out of scope of this PR and the PR itself LGTM.

xuebingli (approved):

app/javascript/pages/admin/IndividualEvents/index.js
samzx: It's just the component state, it isn't shared by anything outside this component so it should be fine I think. Could be a candidate for react hooks :)

PR #192, opened 2019-08-30
Fixed TypeError when setting baseRosettaFile

ThuTrinh:
Fixed TypeError when setting baseRosettaFile

Possibly a side effect from the webpack upgrade. Perhaps the API for compilation.fileDependencies changed from an array to a SortableSet

## Description
When running locally I got the following error:
Screen Shot 2019-08-30 at 10 01 14 am

Looks like `compilation.fileDependencies` is a SortedSet which means we need to use `compilation.fileDependencies.add`

ref: https://codeclimate.com/github/webpack/webpack/lib/util/SortableSet.js/source

## CCs
@zendesk/volunteer


xuebingli (approved):
💯

xuebingli:
@ThuTrinh Thanks for the fix! Could you remove `package-lock.json` before merge, now that we're using Yarn?

samzx:
Thanks for the fix! 😊

PR #193, opened 2019-08-30
Update development readme to reference yarn instead of npm

ThuTrinh:
As part of a webpack update we switched to using yarn, so the development docs needed updating to reflect this.

## CCs
@zendesk/volunteer

## Risks (if any)
Low


xuebingli (approved):
Thanks for the update! How careless of me to forget to do this! 🤦‍♂

PR #194, opened 2019-08-30
Fix uncaught invariant violation console errors

ThuTrinh:
Fix uncaught invariant violation console errors in Events tab while in admin mode

## Description
When loading the Events tab as an admin, uncaught invariant violation console errors are emitted. This is due to the `Cell` delegate of the  `Start` column definition for ReactTable. 

The Cell delegate is required to return either a null or react element. However it was just returning a string. 

Screen Shot 2019-08-30 at 11 38 08 am

The fix was to wrap the string inside a span.

## CCs
@zendesk/volunteer


xuebingli (approved):
👏 Good job, Thu!

Issue #167, opened 2019-06-19
i18n volunteer portal


samzx:
First step, Japanese! 🎉 

~https://github.com/zendesk/client-i18n-tools~
We'll have to use an open source client i18n tool, and write middlewares to make it compatible with existing Zendesk infrastructure to generate the translations. Once generated, we can serve the locales appropriately.

kennethc:
i18n related features are being planned and tracked separately in https://github.com/zendesk/volunteer_portal/projects/15

PR #198, opened 2019-09-11
First cut string extraction

samzx:
First pass to make sure integration with zendesk translation service is working properly.

This required to be merged before: https://github.com/zendesk/rosetta/pull/2975

## Description
We extracted some strings to test functionality. This is the first pass, will mainly go unnoticed to users. If the translation service works, we'll be able to pull the Japanese translations and import it into this repo.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/167)

## CCs
@ovillegaszen 
@zendesk/volunteer 

## Risks (if any)
* low - affects strings


app/javascript/components/Calendar/__snapshots__/test.js.snap
```diff
@@ -1,31 +1,7 @@
...
-    views={
-      Array [
-        "month",
-      ]
-    }
-  />
-</Layout>
+<Consumer>
+  <Component />
+</Consumer>
```
samzx: This probably happened when wrapping the calendar component with `withNamespaces`

app/javascript/components/Calendar/index.js
```diff
@@ -21,7 +23,7 @@ const styles = {
 }
 
 // Custom components given to BigCalendar
-const calendarComponents = (currentUser, offices, filters, filterActions) => {
+const calendarComponents = (currentUser, offices, filters, filterActions, t) => {
```
ThuTrinh: would be good practice to give `t` a more descriptive name for readability.

app/javascript/components/Calendar/index.js
```diff
@@ -145,11 +148,18 @@ const Calendar = ({
         events={selectEvents(events, currentUser, filters)}
         eventPropGetter={eventPropGetter}
         views={['month']}
-        components={calendarComponents(currentUser, offices, filters, {
-          changeShowFilter,
-          changeEventFilter,
-          changeOfficeFilter,
-        })}
+        culture="en"
```
ThuTrinh: is this the culture default value or is it hard coded?

app/javascript/components/Toolbar/index.js
```diff
@@ -94,33 +94,52 @@ const OfficeFilter = ({ value, onChange, offices }) => (
...
       <button className={s.btn} type="button" onClick={() => onNavigate(navigate.NEXT)}>
-        ›
+        {t('dashboard.layoutdatetab.after')}
       </button>
     </div>
     <div className={s.filterBar}>
       <div className={s.filterDropdowns}>
-        <ShowFilter value={showFilter.value} onChange={(_e, _i, value) => showFilter.onChange(value)} />
-        <EventFilter value={eventFilter.value} onChange={(_e, _i, value) => eventFilter.onChange(value)} />
+        <ShowFilter t={t} value={showFilter.value} onChange={(_e, _i, value) => showFilter.onChange(value)} />
```
ThuTrinh: `t` as a component property is not very descriptive. This makes it hard to quickly obtain context or understand how to set this value.

app/javascript/components/Calendar/index.js
samzx: we're just explicitly setting the culture here so in the future we can change it via an interface / menu. But I think "en" is the default

app/javascript/components/Calendar/index.js
samzx: I think it's standard to have `t` as the translation function ie. the API gives the example: https://www.i18next.com/overview/api

What do you think?

ThuTrinh: yeah , I wasn't aware that this prop was added by a third-party. Ignore me :)

ThuTrinh (approved):

ThuTrinh (approved):

PR #204, opened 2019-09-12
[Bug fix] Default translation file and location

samzx:
After, patch up https://github.com/zendesk/rosetta/pull/2975 to use `en` instead of `en-us`

## Description
This will make it compatible with production serving as well as use the default translation name to prevent double fetching.

Also fixed a bug where the compiled json translations would be appending instead of overriding.


xuebingli (approved):
LGTM

PR #205, opened 2019-09-14
Add actions to the default actions export

### LINKED ISSUE #201: Admin Office Filter only displays events in your default office
When new events are created by Admin, the Office filter only shows all events for the specific office setup as user's Default Office

Screen Recording 2019-09-11 at 03 52 pm

kerrisim:
## Description
To fix #201 

The actions were missing from the default actions export so when the default actions object was being imported into the reducer, the reducer could not match on the action being fired. 

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/201)

## Screenshots (if needed)
output
output1

## CCs
@zendesk/volunteer 

## Risks (if any)
* low - reducer actions are being fired now


samzx (approved):
Thanks for this, awesome work 👏

Issue #206, opened 2019-09-25
As an admin when I change the office filter I should see all events for that office


ThuTrinh:
### Steps to reproduce
1. view the list of events in admin mode
2. change the office filter to an office with more events than the one the pages loaded on

### Expected Behaviour
the list of events should show all events for the selected office

### Actual Behaviour
The list of events is restricted to show a max number of events equal to the total events for the office the page originally loaded with.

### Cause
The `ReactTable` component was set with `showPagination={false}` and `defaultPageSize={events.length}`. The component seems to treat these two settings independently and so this combination caused the table to restrict the events shown to the first event list size, but not provide a way to view the next page.

This is a problem if the first list has a small number of events e.g. 0 or 1

![with-default-page-size](https://user-images.githubusercontent.com/3032977/65566885-79ead680-df97-11e9-9330-a24e24060297.gif)

PR #207, opened 2019-09-25
Don't set defaultPageSize when showPagination=false

ThuTrinh:
## Short description
Don't set defaultPageSize when showPagination=false

## Description
The ReactTable component was set with `showPagination={false}` and `defaultPageSize={events.length}`. 
This component seems to treat these two settings independently and so this combination caused the table to restrict the events shown to the first event list size, but not provide a way to view the next page.

This is a problem if the first list has a small number of events e.g. 0 or 1

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/206)

## Screenshots (if needed)
<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

## CCs
@zendesk/volunteer 

## Risks (if any)
* low

### DIFF
--- a/app/javascript/pages/admin/Events/index.js
+++ b/app/javascript/pages/admin/Events/index.js
@@ -119,8 +119,9 @@ const tdProps = () => ({
   },
 })
 
-const Events = ({ data: { networkStatus, events }, deleteEvent }) =>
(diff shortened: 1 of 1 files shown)


xuebingli (approved):
Thanks for the fix, Thu!

PR #212, opened 2019-09-25
Fixed Apollo errors after creating a new event

ThuTrinh:
## Description
1.  Apollo emits errors if you try to read a query that isn't in the cache (matches by query and variables)
- I wrapped the body of the update function in a try/catch block and swallowed the error. We don't really need to do anything if the query is not in the cache. 

2. Apollo needs the query variables to be set when reading and writing to the cache. These things make a unique reference to the query in the cache.
- Updated the code to specify the query variables in both read and write functions.

3. When creating a new event that is not for the current selected office you can't see it in the event list you are redirected towards.
- before redirecting I called the action to set the officeFilter to match that of the newly created event 

## References
None

## Screenshots (if needed)
<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

## CCs
@zendesk/volunteer 

## Risks (if any)
* Low

### DIFF
--- a/app/javascript/pages/admin/Events/index.js
+++ b/app/javascript/pages/admin/Events/index.js
@@ -173,20 +173,22 @@ const withData = compose(
           variables: { id: event.id },
           optimisticResponse: buildOptimisticResponse(event),
           update: (cache, { data: { deleteEvent } }) => {
-            const queryVariables = {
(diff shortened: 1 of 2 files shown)


Pancrisp (approved):
Thanks for fixing this!

Issue #53, opened 2018-08-01
Fix Top Volunteers on Dashboard


danielbreves:
### References
* Ticket 76

danielbreves:
It seems like this query is still not returning sorted results: https://github.com/zendesk/volunteer_portal/blob/4963611e1590aafbd8a57fec4b48c024aba1ae7c/app/graph/resolvers/user_resolver.rb#L85-L115. It may still not be completely accurate.

timbertson-zd:
We came up with a modified SQL that works, but haven't started the work to represent it in activerecord yet:

```
SELECT
  "users".id, "users".first_name, users.last_name,
  event_durations.total, individual_event_durations.total,
  COALESCE(individual_event_durations.total,0) + COALESCE(event_durations.total,0) as total_duration
  from users
  left outer join (
    select users.id as user_id, 
    SUM((EXTRACT(EPOCH FROM "events"."ends_at") - EXTRACT(EPOCH FROM "events"."starts_at")) / 60) as total
    FROM "users" 
    INNER JOIN "signups" ON "signups"."user_id" = "users"."id" 
    INNER JOIN "events" ON "events"."id" = "signups"."event_id" 
      AND ("events"."starts_at" > '2017-12-31 13:00:00') 
      AND ("events"."starts_at" < '2018-12-31 13:00:00') 
    GROUP BY users.id
  ) as event_durations
  on event_durations.user_id = users.id

  left outer join (
    select users.id as user_id, 
    SUM("individual_events"."duration") as total
    FROM "users" 
    INNER JOIN "individual_events" ON "individual_events"."user_id" = "users"."id" 
      AND "individual_events"."date" > '2018-01-01' 
      AND "individual_events"."date" < '2019-01-01' 
    GROUP BY "users"."id"
  ) as individual_event_durations
  ON event_durations.user_id = individual_event_durations.user_id
  and users.id = event_durations.user_id

WHERE "users"."deleted_at" IS NULL
ORDER BY total_duration
;
```

The basic idea is two independent queries where you get one row per user (even if the duration is 0), and then join them on user.id and sum the two types of durations.

timbertson-zd:
(we also noticed that the start / end conditions are a little inconsistent between the two types, which is another bug)

screamish:
From a discussion with @xuebingli @kennethc just now:

[...]

PR #213, opened 2019-10-09
Fix top volunteers ranking in the dashboard

kennethc:
Fix top volunteers ranking in the dashboard

## Description
The UserResolver, when returning a list of top volunteers, will join the `events` and `individual_events` tables, then calculate the sum of the durations for each user to give you a ranking.

This join operation causes duplication of events or individual events, which means that a user might be reported with more volunteering hours than they should have. This causes the list to sort the users in the incorrect order.

A join operation is used to combine another table which contains additional columns for the first table. However, `individual_events` actually is additional rows to the `events` table and not additional columns. A union operation is the appropriate operation to combine additional rows from another table.

As this is a rather big change, a new resolver is created to limit potential impact to other parts of the application.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/53)

## Tasks
- [x] Write tests

## CCs

If app migration related
@zendesk/volunteer 

## Risks (if any)
* medium - Top Volunteers in the dashboard may not be rendered correctly


app/graphql/resolvers/top_volunteer_resolver.rb
```diff
@@ -0,0 +1,79 @@
...
+      events_scope, individual_events_scope = scope_with_time(events_scope, individual_events_scope, args[:after], args[:before])
+      events_scope, individual_events_scope = scope_with_office_id(events_scope, individual_events_scope, office_id)
+
+      all_events_scope = events_scope.union_all(individual_events_scope)
+      all_events_scope = all_events_scope.select("#{user_entry},  SUM(\"duration\") AS \"duration\"").group(user_entry)
+
+      all_events_scope = scope_with_count(all_events_scope, args[:count])
+      all_events_scope = scope_with_sort_by(all_events_scope, args[:sort_by])
+
+      all_events_scope
```
kennethc: If all the scopes are applied, this will look something like:

```
SELECT  "users"."id", "users"."first_name", "users"."last_name", "users"."email", "users"."group", "users"."photo",  SUM("duration") AS "duration" 
FROM (
  (
    SELECT "users"."id", "users"."first_name", "users"."last_name", "users"."email", "users"."group", "users"."photo", "events"."duration" 
    FROM "users" 
    INNER JOIN "signups" ON "signups"."user_id" = "users"."id" 
    INNER JOIN "events" ON "events"."id" = "signups"."event_id" 
    WHERE "users"."deleted_at" IS NULL 
    AND ("events"."starts_at" > '2018-12-31 13:00:00') 
    AND ("events"."starts_at" < '2019-10-09 01:38:01') 
    AND "users"."office_id" = $1
  ) 
  UNION ALL 
  (
    SELECT "users"."id", "users"."first_name", "users"."last_name", "users"."email", "users"."group", "users"."photo", "individual_events"."duration" 
    FROM "users" 
    INNER JOIN "individual_events" ON "individual_events"."user_id" = "users"."id" 
    AND "individual_events"."deleted_at" IS NULL 
    WHERE "users"."deleted_at" IS NULL 
    AND ("individual_events"."date" > '2018-12-31 13:00:00') 
    AND ("individual_events"."date" < '2019-10-09 01:38:01') 
    AND "users"."office_id" = $2
  ) 
) "users" 
GROUP BY "users"."id", "users"."first_name", "users"."last_name", "users"."email", "users"."group", "users"."photo" 
ORDER BY "duration" DESC 
LIMIT $3
```

app/graphql/resolvers/top_volunteer_resolver.rb
```diff
@@ -0,0 +1,79 @@
+module TopVolunteerResolver
+  class << self
+    def all(_object, args, context)
+      user_entry = '"users"."id", "users"."first_name", "users"."last_name", "users"."email", "users"."group", "users"."photo"'
+
+      events_scope = User.joins(:events).select("#{user_entry}, \"events\".\"duration\"")
```
ThuTrinh: is there a reason why we are using string interpolation here?

app/graphql/resolvers/top_volunteer_resolver.rb
```diff
@@ -0,0 +1,79 @@
+module TopVolunteerResolver
```
ThuTrinh: Typically a resolver would be responsible for a single resource/query type, and the use of query params would control the filtering of data.

In this case I would suggest calling the Module `VolunteerResolver` and add orderby and pagination. This will support getting the top N volunteers and much more.

app/graphql/resolvers/top_volunteer_resolver.rb
kennethc: `user_entry` is used in multiple locations to denote the fields we actually want from `User`. Instead of repeating the same user fields in different parts of the code, putting it in a variable and using string interpolation reduces the duplication of the string literals that we'd otherwise have.

app/graphql/resolvers/top_volunteer_resolver.rb
samzx: Good point, we added some comments to add some context of this existence and also will change the name!

app/graphql/resolvers/top_volunteer_resolver.rb
kennethc: Did you actually mean to ask why we're not using something like: `User.joins(:events).select(:id, :first_name)`?

That's a good point. I wasn't too aware of the usage of symbols here. It looks like it does work, but the last `:duration` would not translate to `"events"."duration"` in SQL.

Right now it will translate to just `"duration"`, which can be ambiguous. How would I reference it so it translates nicely to `"events"."duration"`?

kennethc: I answered my own question. Just need to use `:'events.duration'`.

kennethc: Bumped into a problem further down and I don't know how to solve it.
`all_events_scope = all_events_scope.select("#{user_entry},  SUM(\"duration\") AS \"duration\"").group(user_entry)`

I'm not sure how we can convert the `SUM()` here into something that is compatible with symbols.

test/graphql/resolvers/volunteer_resolver_test.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      results.must_equal [user]
+    end
+
+    it 'filters by office_id' do
+      args = { office_id: office.id }
+      results = VolunteerResolver.all(nil, args, nil).to_a
+
+      results.must_equal [user]
+    end
```
kennethc: These tests are failing. I believe there is some setup being done somewhere else for `UserResolver` which we have to replicate for `VolunteerResolver`.

kennethc: Actually, scrap the above, I just figured out why these don't work.

`UserResolver` does not rely on the user being signed up to an event to return a list of users, while `VolunteerResolver` does. This resolver was created to solve the problem that the top volunteer dashboard returns an incorrect list of top volunteers (hence the previous name). So it makes no sense to query for users without events.

app/graphql/resolvers/top_volunteer_resolver.rb
kennethc: Just want to provide a bit more clarity on this topic. This resolver was created to return a list of users ordered by volunteer hours. Unlike the `UserResolver`, this resolver doesn't just return a list of users. Instead, it tries to expose a concept of "user-hours" based on users and events. This concept isn't well represented in code but is used in a couple of places like the top volunteers dashboard and the admin reporting function.

kennethc:
This PR is ready for final review.

@ThuTrinh, I would appreciate your feedback and 👍 here.

test/graphql/resolvers/volunteer_resolver_test.rb
```diff
@@ -0,0 +1,163 @@
...
+      args = { office_id: office.id }
+
+      Signup.create!(event: event1, user: user)
+      results = VolunteerResolver.all(nil, args, nil).to_a
+
+      results.must_equal [user]
+    end
+
+    it 'limits to count' do
+      args = { count: 0 }
```
ThuTrinh: This test could generate a false positive if there was something broken. Would be better to test both `0` and `n` case.

app/graphql/resolvers/top_volunteer_resolver.rb
ThuTrinh: yeah ok, I am not familiar with active record, so I can't provide in-depth help here sorry. I was just worried about opening holes for sql injection. However after our chat, you have convinced me that the parts that are not using symbols are in our control.

ThuTrinh (approved):
Thanks for doing this @kennethc , this was a big effort.

samzx (approved):
Bravo 👏🙂

PR #215, opened 2019-10-09
Support creating directories for node versions below v10

samzx:
# Description
NodeJS version 10.12.0 has added a native support for both mkdir and mkdirSync to create a directory recursively with `recursive: true` option.

For users with node versions below, we'll need to do it this way.

## Risks (if any)
low


KaiminHuang (approved):
Cool, I had the same problem during setup. 
Good to go on 🍏

sunstryder (approved):

sunstryder:
Looks good. I think we should probably update our node version to at least 10.16

PR #217, opened 2019-10-09
Add architecture diagram to development docs

ThuTrinh:
Add architecture diagram to development docs

## CC
@zendesk/volunteer


docs/development.md
```diff
@@ -9,14 +13,14 @@
    cd volunteer_portal
    ```
 
-1. Install the right version of ruby and node
+2. Install the right version of ruby and node
```
xuebingli: nit: this is actually intentional. Always use `1` make (re)ordering, inserting, and removing much easier.

ThuTrinh: I am confused, why do we use numbered list then? why not just a bullet list?

Pancrisp: GitHub knows when to adjust numbered lists accordingly even if they're all prefixed with `1.` in the Markdown file. What @xuebingli is saying is that you won't have to change the numbers when you reorder your lists because they're all 1s, GitHub will automatically display them in the correct sequence.

samzx (approved):

PR #220, opened 2019-10-09
Fix security vulnerabilities for js-yaml

Pancrisp:
## Description
Migrated `css-loader` to v3 
* Removed minimize option in webpack config for `css-loader` (minification no longer supported in v1 and beyond)
* Added `cssnano` postcss plugin as replacement for minifying CSS

Upgraded `js-yaml` to 3.13.1

## Risks (if any)
* **Low**: Major version bump for `css-loader` and the addition of cssnano as a postcss plugin could break webpack bundling.


xuebingli (approved):
LGTM 👏

Issue #227, opened 2019-10-23
Investigate/Fix Jest Snapshot tests


ThuTrinh:
We use Jest to perform snapshot tests in our CI pipeline. When we update our view components we need to also update our snapshots locally and commit them.

Jest seems to generate different snapshots when running locally compared to that in CI. This causes failing test and blocks merging.

Perhaps look into how Jest generates these snapshots. There might be some settings that differ between environments.

samzx:
https://github.com/zendesk/volunteer_portal/pull/231 closes this

PR #231, opened 2019-10-23
Fix Jest tests on circleci

samzx:
Update snapshots with new cached dependancies.
Previously, the jest tests would be passing on outdated node_modules that were cached incorrectly. We were check summing on the yarn file while using npm.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,17 +65,16 @@ jobs:
       - image: circleci/node:8-stretch-browsers
     steps:
       - checkout
-      # - restore_cache:
-      #     keys:
-      #       # when lock file changes, use increasingly general patterns to restore cache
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -67,19 +67,17 @@ jobs:
...
       - restore_cache:
           keys:
-          - v1-js-dependencies-{{ checksum "yarn.lock" }}
-          # fallback to using the latest cache if no exact match is found
-          - v1-js-dependencies-
+            - yarn-packages-v2-{{ checksum "yarn.lock" }}
       - run:
           name: install dependencies
-          command: npm install
+          command: yarn install
```
sunstryder: yep that'll do

sunstryder (approved):

samzx:
closes #227

Issue #235, opened 2019-10-24
Get JS packages up to date


kayison:
A number of packages are out of date. Some of these packages, such as `react-router-dom` are a few major versions behind. This has prevented other upgrades such as `react` v16.

Some others are deprecated or not needed.

Get packages up to date and then look at automating this with (dependabot perhaps because it's also compatible with the Ruby gems) #233

PR #236, opened 2019-10-24
Bump babel packages

kayison:
Minor bump babel packages and delete unused

## Description
Part of a series of PRs to get npm packages up to date.
- This updates babel packages (minor and patch)
-  Remove unused babel proposals. These should only be added if/when we are using those features. These won't compile with babel if they are used in the code but not needed - so safely removed
- Remove `plugin-syntax-object-rest-spread` as the functionality is covered by `plugin-proposal-object-rest-spread`

## References
[Github Issue #235](https://github.com/zendesk/volunteer_portal/issues/235)

## Risks
* Low: This should have compile issues if something is not going to work


Pancrisp (approved):
Thanks for kicking this off @kayison! Our dependencies are in dire need of pruning.

PR #238, opened 2019-11-06
Fix missing calendar events for events in the same office

samzx:
When the calendar is first loaded, events are filtered though the `applyOfficeFilter` code path.


kayison (approved):
`event.office.id` will blow up if it doesn't exist but I've been told that users are assigned a default office

👍

Issue #224, opened 2019-10-13
Triple avatar when signing up bug


samzx:
When signing up to events, the UI make three avatars appear to have signed up

> I signed up for an event and it decided I was worth 3 regular volunteers :wink:
(Refreshing showed correct results, so presumably just a client-side bug)

> still seeing a triple sign up happen when I click ‘join’ on an event. It’s not signing me up three times, just bizarre UI behaviour. After this all events dispeared from the month I was viewing too. I had to do a refresh

References (Zendesk Only):
https://zendesk.slack.com/archives/GBTLTQLQN/p1569216217001400
https://zendesk.slack.com/archives/GBTLTQLQN/p1571000067000100

samzx:
another report:
https://zendesk.slack.com/archives/GBTLTQLQN/p1572827398000100

kayison:
This bug was introduced when `inverse_by` was added to https://github.com/zendesk/volunteer_portal/blob/8ee7427c71ea3df350656b8b32bbb8156737c03f/app/models/event.rb#L4

I'm not exactly sure why it's causing the duplicates yet though

PR #240, opened 2019-11-06
Fix duplicate avatars when signing up

### LINKED ISSUE #45: As an admin, I want to see user details in case of emergencies
Would be good to come with a way that ensure these are kept up to date.

kayison:
## Description
This is a temporary fix. `inverse_by` was introduced in https://github.com/zendesk/volunteer_portal/commit/8ee7427c71ea3df350656b8b32bbb8156737c03f due to rubocop errors.
When `inverse_by` is on `signing` in `Events`, it causes three Avatars when signing up to an event. This fixes it for now but further investigation should happen to determine why this happens.
Fixes #45

## References
[Github Issue #224](https://github.com/zendesk/volunteer_portal/issues/224)

## Risks (if any)
* Low: fix bug, removing `inverse_of` unlikely to have terrible effects here

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -1,7 +1,11 @@
 class Event < ApplicationRecord
   CALENDAR_TIME_FORMAT = '%Y%m%dT%H%M%SZ'.freeze
 
-  has_many :signups, -> { auto_include(false) }, dependent: :destroy, inverse_of: :event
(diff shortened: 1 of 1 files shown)


samzx (approved):
👍 thanks for the fix! Would like to know the cause of this too

PR #241, opened 2019-11-06
Update Header to use Garden & styled components w/ Language controls

### LINKED ISSUE #209: Menu/option to select language
We'd like either an auto-detect or menu somewhere where we could select the language. Following #167

samzx:
## Description
Updated header to use the modern stack: garden components and using styled components. Primary reason is to add language controls (allowing user to manually change language settings).

There will be follow up PRs to address auto-detection of language.

We've also moved the admin navigation into the user profile space. It also dynamically allows a link back to the volunteer interface if you're in the admin interface.

closes #209 

## Screenshots (if needed)
image

image

## CCs

@ovillegaszen


ThuTrinh (approved):

ThuTrinh:
@samzx - This branch will give me conflicts in mine. I am keen to get this merged so I can resolve my Godzilla branch :trollface:

__mocks__/fileMock.js
```diff
@@ -0,0 +1 @@
+module.exports = 'test-file-stub'
```
ThuTrinh: I am not too sure about this location. It's really only relevant to the js tests. I would rather just put it in `/volunteer_portal/app/javascript/test/__mocks__/fileMock.js`

ThuTrinh:
hmmm, do you think it would be more consistent if the language selection and admin toggle were in the profile menu?

__mocks__/fileMock.js
samzx: good point, I'll just have to update the jest config first (by default it points to this directive) :)

samzx:
> hmmm, do you think it would be more consistent if the language selection and admin toggle were in the profile menu?

I was thinking about this, and I think that's what we'd move towards, but we'll need to refactor the the profile menu to garden components.

We can probably do it in this PR (put things in the menu), and let you merge your branch first - the translations aren't quite ready yet anyway.

app/javascript/components/Header/index.js
```diff
@@ -1,40 +1,237 @@
...
 import { UserContext, FilterContext } from '/context'
-import { present } from '../../lib/utils'
-import s from './main.css'
 
-const styles = {
-  item: {
-    fontSize: 14,
-  },
+const NavigationText = styled.div`
+  color: #08c;
```
ThuTrinh, line 26: can we use garden variables for the colours?

e.g.
```
const { zdColorGrey600, zdColorSecondaryAzure600 } = require('@zendeskgarden/css-variables')

const StyledSvgAvatar = styled(GardenAvatar)`
  background-color: ${zdColorGrey600};
`
```

app/javascript/components/Header/index.js
```diff
@@ -1,40 +1,237 @@
...
+`
+
+const UserDetails = styled.div`
+  margin-left: 10px;
```
ThuTrinh, line 30: can we use the spacing in the theme here please?
e.g.
```
const spacingXxs = R.path(['theme', 'styles', 'spacing', 'xxs'])
const UserDetails = styled.div`
  margin-left: 10px; ${spacingXxs};
  }
`
```

ThuTrinh (commented):
I need to go walk my dogs, so I have not finished reviewing yet. But here are a couple of minor things to start with.

app/javascript/components/Header/index.js
```diff
@@ -1,40 +1,237 @@
...
+  display: flex;
+  border: none;
+  text-align: left;
+`
+
+const UserName = styled(MD)`
+  font-weight: bold;
+`
+
+const UserProfile = ({ currentUser, offices, handleOfficeSelect, location, router }) => {
```
ThuTrinh, line 43: There is a lot going on in this file. It is actually housing 2 separate components. I would recommend moving the `UserProfile` component into it's own file for easy discovery and better separation of concerns.

In the `Header` component, have you considered using `{children}` instead of explicitly using `UserProfile`? 

The two components would be more composable after these suggestions imo.

app/javascript/components/Header/index.js
```diff
@@ -1,40 +1,237 @@
...
+    } else if (selectedItem === 'admin') {
+      router.push('/portal/admin')
+    } else if (selectedItem === 'home') {
+      router.push('/portal')
+    } else if (selectedItem === 'sign-out') {
+      router.push('/portal/sign_out')
+    }
+  }
+
+  const renderItems = () => {
```
ThuTrinh, line 102: There is quite a lot going on in this function. Separating sub components out into their own files under the same folder can help readability and separation of concerns.
i.e. The UserProfile component could just care about what to show and not also how to show each thing in the menu.

app/javascript/components/Header/index.js
```diff
@@ -1,40 +1,237 @@
...
+  const [tempSelectedItem, setTempSelectedItem] = useState()
+  const [selectedItem, setSelectedItem] = useState({ office: currentUser.office, language: languages[0] })
+
+  const handleStateChange = (changes, stateAndHelpers) => {
+    if (R.hasPath(['isOpen'])(changes)) {
+      setIsOpen(
+        changes.selectedItem === 'default-office' ||
+          changes.selectedItem === 'language-settings' ||
+          changes.selectedItem === 'general-settings' ||
+          changes.isOpen
```
ThuTrinh, line 61: Can we not just check if `changes.selectedItem` has a value? when would it be set to anything but these three?

samzx: changes.selectedItem also could be a menu item, ie. 'English' (an item of languages) in which we would want to close the menu, as opposed selecting a sub-menu.

app/javascript/components/Header/index.js
```diff
@@ -1,110 +1,55 @@
...
+
+const Logo = styled.img`
+  height: 40px;
+`
+
+const LogoBox = styled.div`
+  flex-grow: 2;
+`
+
+const LogoSection = () => (
```
ThuTrinh, line 32: this is not being used anywhere else but inside `Header`, I would just embed it inside tbh.

app/javascript/components/UserProfileMenu/Menus.jsx
```diff
@@ -0,0 +1,69 @@
...
+import {
+  Item,
+  Separator,
+  PreviousItem,
+  HeaderItem,
+  NextItem,
+} from '@zendeskgarden/react-dropdowns'
+import TranslationIcon from '@zendeskgarden/svg-icons/src/12/translation-exists-fill.svg'
+
+export const DefaultOfficeMenu = ({ offices, previousMenuValue, selectedItem, t }) =>
```
ThuTrinh, line 13: Typically each file would contain one component, I would have expected three different files here.

app/javascript/components/UserProfileMenu/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+const UserName = styled(MD)`
+  font-weight: bold;
+`
+
+const UserProfileMenu = ({ offices, location, router, togglePopover }) => {
+  const { currentUser, setOffice } = useContext(UserContext)
+  const { setOfficeValue } = useContext(FilterContext)
+  const [ updateDefaultOffice ] = useMutation(UpdateUserOfficeMutation)
+
+  if (R.isNil(currentUser) || R.isEmpty(currentUser)) return null
```
ThuTrinh, line 39: did you get a warning? I don't think this fast return can go here because it creates a situation where the same number of hooks are not applied in every render. Hooks need to be applied in the same order as well.

[reference](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

app/javascript/components/UserProfileMenu/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+      }
+    } else if (tempSelectedItem === 'language-settings') {
+      if (R.hasPath(['language'])(selectedItem)) {
+        i18n.changeLanguage(selectedItem.language.value, () => {
+          // TODO: Handle callback (error/success)
+        })
+        setSelectedItem(selectedItem)
+      }
+    } else if (selectedItem === 'admin') {
+      router.push('/portal/admin')
```
ThuTrinh, line 98: if the menu items themselves are router links, do we need to do this here?

app/javascript/components/Header/index.js
samzx: good point!

app/javascript/components/UserProfileMenu/index.jsx
```diff
@@ -0,0 +1,144 @@
...
+          <Avatar>
+            <img src={currentUser.photo} alt="User Avatar" />
+          </Avatar>
+          <UserDetails>
+            <UserName>{currentUser.name}</UserName>
+            <MD>{currentUser.office && currentUser.office.name}</MD>
+          </UserDetails>
+        </UserProfileContainer>
+      </Trigger>
+      <GardenMenu hasArrow>{renderItems()}</GardenMenu>
```
ThuTrinh, line 139: you don't need to change it, but you could use ramda here:
```
<GardenMenu hasArrow>{
        R.cond([
          [R.equals('default-office'). R.always(<DefaultOfficeMenu offices={offices} previousMenuValue="general-settings" selectedItem={selectedItem} t={t} />)],
          [R.equals('language-settings'), R.always(<LanguageMenu languages={languages} previousMenuValue="general-settings" selectedItem={selectedItem} t={t} />)],
          [R.T, R.always(<GeneralSettingsMenu menuValues={{ defaultOffice: 'default-office', languageSettings: 'language-settings'}} isAdmin={currentUser.isAdmin} pathname={location.pathname} t={t} />)]
        ])(tempSelectedItem)
      }
</GardenMenu>
```

app/javascript/components/UserProfileMenu/index.jsx
samzx: No warning, this was just ported from the header section (which didn't display if there was no user). Would scoping it to just the `UserProfileContainer` be better?

app/javascript/components/UserProfileMenu/Menus.jsx
samzx: Okay I will do that, I was lazy 😛

app/javascript/components/UserProfileMenu/index.jsx
samzx: it's for accessibility - since when a user selects this item (via tabbing etc), it will trigger the select handler instead of the Link

I'll add a comment for this

app/javascript/components/UserProfileMenu/index.jsx
samzx: might be great to use this if we needed to compose the boolean condition with something more complex 👍

app/javascript/components/UserProfileMenu/index.jsx
ThuTrinh: hmm weird. I would prefer all the hooks to be grouped up at the top though please. for consistency and readability.

ThuTrinh (approved):

PR #250, opened 2019-11-20
Update pageSize based on number of users

### LINKED ISSUE #248: Fix admin export office filter
There is some behaviour where the filter behaviour is either:
- not showing everyone
- office filter exporting not based on the filter, ie. setting office to all, sometimes you'd only get the export for melbourne office.

samzx:
Previously it doesn't update the size of the table if the number of users in the table change. This fixes that. For example, if there are 50 users loaded, the table will only show 50 users. If an admin changes the filters to "all" and there are say 100 users, still, only 50 users are shown.

Follow up:

- [ ]  Add a loading indicator to signal that it's fetching user data.

closes #248

### DIFF
--- a/app/javascript/components/Reporting/index.js
+++ b/app/javascript/components/Reporting/index.js
@@ -146,6 +146,7 @@ const Reporting = ({ users, startDate, endDate, onStartChange, onEndChange }) =>
     data={users}
     showPagination={false}
     defaultPageSize={users.length}
+    pageSize={users.length}
     minRows={0}
     defaultFilterMethod={defaultFilterMethod}
     getProps={containerProps}


kennethc (approved):
Looks like it fixes the immediate problem but could be a usability nightmare.

jliclaire (approved):
🍏

samzx:
> Looks like it fixes the immediate problem but could be a usability nightmare.

Yes, we'd add pagination as a follow up. Though pagination is quite difficult since the export function is tied to visible rows and exporting all user data will require a filter to pass through it.

kennethc:
> > Looks like it fixes the immediate problem but could be a usability nightmare.
> 
> Yes, we'd add pagination as a follow up. Though pagination is quite difficult since the export function is tied to visible rows and exporting all user data will require a filter to pass through it.

Maybe we can consider a reporting API that will give you that report, instead of exporting what's on the screen?

samzx:
> Maybe we can consider a reporting API that will give you that report, instead of exporting what's on the screen?

Yep, good idea, that's what we'd likely want to move towards in the future :)

PR #252, opened 2019-11-21
Package upgrades get ready for React v16

kayison:
## Description
- Get some packages in a version that allows for both React v15 and v16 in preparation for upgrading to React v16. This allows us to upgrade to React v16 (and its friends `react-dom` etc)  in a single PR without the other packages.
- This is going to be a multi-step process because some of these packages are so out of date (and React v16 has been out for so long) that the latest versions of those packages don't allow React v15. 
- This PR is to ensure we upgrade incrementally and don't put everything (and the code fixes for breaking changes) in one giant commit. 

- `react-google-maps` is still in a React v15 version because the version that is compatible with both v15 and v16 should be in its own PR due to the large amount of breaking changes. It is also a deprecated package and should be moved to something else. 

- The bump to `material-ui` has changed a few css variables that don't appear to have affected much. 

## References
[Github Issue # 235](https://github.com/zendesk/volunteer_portal/issues/235)

## Risks
* Medium: Chance that some of the packages have breaking changes that haven't been picked up. Needs to be thoroughly tested.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,7 @@
     "ramda": "^0.23.0",
     "react": "^15.6.2",
     "react-apollo": "^2.0.4",
-    "react-big-calendar": "^0.12.3",
+    "react-big-calendar": "0.17.0",
     "react-dom": "^15.6.2",
     "react-geosuggest": "^2.7.0",
     "react-google-maps": "^4.11.0",
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
     "react-router": "^3.2.5",
     "react-router-redux": "^4.0.5",
     "react-table": "^6.7.6",
-    "react-tap-event-plugin": "^2.0.1",
     "redux": "^3.5.2",
     "redux-form": "^7.2.0",
     "redux-logger": "^2.6.1",
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "path-complete-extname": "^0.1.0",
     "postcss-loader": "^2.0.10",
     "postcss-smart-import": "^0.7.6",
-    "prop-types": "^15.6.0",
+    "prop-types": "^15.7.2",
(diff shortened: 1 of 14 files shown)


ThuTrinh (approved):

package.json
```diff
@@ -79,27 +79,26 @@
...
     "react-apollo": "^2.0.4",
-    "react-big-calendar": "^0.12.3",
+    "react-big-calendar": "0.17.0",
     "react-dom": "^15.6.2",
     "react-geosuggest": "^2.7.0",
     "react-google-maps": "^4.11.0",
     "react-i18next": "^8.3.9",
     "react-redux": "^4.4.5",
-    "react-router": "^2.6.0",
+    "react-router": "^3.2.5",
```
Pancrisp: Would there be a fix for breaking changes migrating to v3?

The save button does not work when adding new events in the admin view when using v3 (http://localhost:5000/portal/admin/events/new)

The events still get created, but the page does not get redirected to the dashboard.

samzx: is it that `props.history.push` is deprecated in favour of using the new `this.context.history`

Pancrisp: `history.push` is still valid in React Router v5.

kayison: Thanks for that @Pancrisp. I had heard there was something that broke in v3 in the app but wasn't sure which. And yeah the changelog when I looked didn't identify anything suspicious. 

I'll hold off on merging this PR and do some debugging when I'm working on this next Wednesday.

kayison: I've modified `props.history.push` to `props.router.push` and it appears to be working now.

sunstryder: great news :)

samzx (approved):
LGTM 🎉

sunstryder (approved):

Pancrisp (approved):

Issue #256, opened 2019-12-04
Enable IP safe-listing and deny-listing using Rack attack


samzx:
We'd like to be able to add a feature where we can Allow and Block traffic from certain IP addresses. This feature can be used to only allow connections from VPN connections (if you know the IP addresses).

The Middleware of choice would likely be https://github.com/kickstarter/rack-attack

samzx:
solved by #258

PR #258, opened 2019-12-04
Add ability to allow / block ip addresses

samzx:
Add ability to allow / block ip addresses

## Description
This allows deployers to only administer the volunteer portal to certain IP addresses, such as addresses under a certain VPN connection to control data access.

The environment variables should be configured to allow this behaviour.

Solves #256 

## CCs
@zendesk/volunteer


kayison (approved):

ThuTrinh (approved):

ThuTrinh:
@samzx - why did we move this back into backlog?

samzx:
> @samzx - why did we move this back into backlog?

Pending security reply, but we can merge this anyway since it's optional to use.

PR #266, opened 2019-12-18
Upgrade to react v16

kayison:
## Description
- Upgrade to React v16 so we can use libraries that make use of React v16 as well as use functionality such as hooks.
- Move to React 16 enzyme adapter for tests (the number matches the React version).
- Remove PropTypes that was using it from React, as it is in a separate package now.
- In `Event` there is a `componentWillMount` method which is deprecated due to being unsafe. It has been changed to a `componentDidMount` which shouldn't change its current functionality too much (it already does a double render anyway with the old method and should eventually get fixed up. 

## References
[Github Issue #235 ](https://github.com/zendesk/volunteer_portal/issues/235)

## Risks (if any)
* Low: React is pretty backwards compatible, but we should check to make sure everything works as expected


kennethc (approved):

kayison:
Closed (and reopened by @samzx) due to CircleCI not triggering

PR #271, opened 2019-12-20
Exclude css-module for @zendeskgarden scope

samzx:
The `modules` in the `css-loader` webpack plugin hashes the garden component classnames into those random strings. When you use garden components they have a specific class name like "im-a-tag". When you import `@zendeskgarden/react-tags/dist/styles.css` , everything get's hashed - so "im-a-tag" becomes "sdnKDSnd".

We can turn off the hashing for everything inside the garden path so they will work now :)

Thanks @kayison for guiding me to the solution 😀


kayison (approved):
👍

PR #273, opened 2019-12-27
Fix Autocomplete for Organizations Field (after React 16 upgrade)

samzx:
The Boxes are kind of mis-aligned but that's alright?

## CCs

If app migration related
@zendesk/volunteer

### DIFF
--- a/app/javascript/components/LocationField/index.js
+++ b/app/javascript/components/LocationField/index.js
@@ -10,7 +10,13 @@ const LocationField = ({ input: { value, onChange } }) => (
     suggestsClassName={s.autocompleteList}
     initialValue={value}
     onChange={onChange}
-    onSuggestSelect={s => onChange(s.gmaps.formatted_address)}
+    onSuggestSelect={s => {
+      try {
+        onChange(s.gmaps.formatted_address)
+      } catch {
+        onChange('')
+      }
+    }}
   />
 )
 
--- a/app/javascript/components/LocationField/index.js
+++ b/app/javascript/components/LocationField/index.js
@@ -1,6 +1,8 @@
 import React from 'react'
 import Geosuggest from 'react-geosuggest'
 
+import R from 'ramda'
+
 import s from './main.css'
 
 const LocationField = ({ input: { value, onChange } }) => (
@@ -10,13 +12,10 @@ const LocationField = ({ input: { value, onChange } }) => (
     suggestsClassName={s.autocompleteList}
     initialValue={value}
     onChange={onChange}
-    onSuggestSelect={s => {
-      try {
-        onChange(s.gmaps.formatted_address)
-      } catch {
-        onChange('')
-      }
-    }}
+    onSuggestSelect={R.pipe(
+      R.pathOr('', ['gmaps', 'formatted_address']),
+      onChange
+    )}
   />
 )
 
--- a/app/javascript/components/ReduxFormAutoComplete/index.js
+++ b/app/javascript/components/ReduxFormAutoComplete/index.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import { Dropdown, Menu, Item, Autocomplete, Field as GardenField } from '@zendeskgarden/react-dropdowns'
 import debounce from 'lodash.debounce'
+import R from 'ramda'
 
 const ReduxFormAutocomplete = ({ dataSource, input: { onChange }, searchField }) => {
   const searchData = dataSource.map(data => data[searchField])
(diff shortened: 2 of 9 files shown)


app/javascript/components/ReduxFormAutoComplete/index.js
```diff
@@ -0,0 +1,70 @@
...
+  const searchData = dataSource.map(data => data[searchField])
+  // Selected Item is the String
+  const [selectedItem, setSelectedItem] = React.useState('')
+  const [inputValue, setInputValue] = React.useState('')
+  const [matchingOptions, setMatchingOptions] = React.useState(searchData)
+
+  /**
+   * Debounce filtering
+   */
+  const filterMatchingOptionsRef = React.useRef(
```
ThuTrinh: why are we using a `useRef` hook here? afaik the useRef hook here will create a mutable variable for the lifetime of this component. I don't see it being mutated anywhere and it is also a const.

app/javascript/components/ReduxFormAutoComplete/index.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  return (
+    <Dropdown
+      inputValue={inputValue}
+      selectedItem={selectedItem}
+      onSelect={item => {
+        const itemData = dataSource.find(data => data[searchField] === item)
+        setSelectedItem(item)
+        onChange(itemData.id)
```
ThuTrinh: what if `data[searchField] !== item`  `and itemData` is undefined? `onChange(itemData.id)` would throw a null ref exception I think. 
You can use Ramda to safely extract the id

```
onChange(R.prop('id', itemData))
```

app/javascript/components/ReduxFormAutoComplete/index.js
samzx: I think this was part of the React Garden examples, and the variable is mutated when `inputValue` is changed in the `useEffect` block. The `const` might be so it can't be re-assigned?

app/javascript/components/LocationField/index.js
```diff
@@ -10,7 +10,13 @@ const LocationField = ({ input: { value, onChange } }) => (
     suggestsClassName={s.autocompleteList}
     initialValue={value}
     onChange={onChange}
-    onSuggestSelect={s => onChange(s.gmaps.formatted_address)}
+    onSuggestSelect={s => {
+      try {
+        onChange(s.gmaps.formatted_address)
```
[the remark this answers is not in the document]
samzx: oh that is super neat 😍

app/javascript/components/ReduxFormAutoComplete/index.js
samzx: If it does happen, I suppose assigning it to `undefined` (no selection) is better than crashing the app, I like this, Ramda is so nice 😃

app/javascript/components/ReduxFormAutoComplete/index.js
samzx: For reference: https://garden.zendesk.com/react-components/dropdowns/#!/Autocomplete%20usage

ThuTrinh: From my understanding `debounce` will return a function `A`, and `useRef` here will return a mutable ref object whose .current property is initialized to `A`. So in the `useEffect` block, we are essentially just executing function `A` right?

I just don't see the point of using `useRef` here, unless I am missing something. 

e.g. you could do the same thing by just removing it:

```
const filterMatchingOptionsRef = debounce(value => {
      const matchingOptions = searchData.filter(data => {
        return (
          data
            .trim()
            .toLowerCase()
            .indexOf(value.trim().toLowerCase()) !== -1
        )
      })

      setMatchingOptions(matchingOptions)
    }, 300)

React.useEffect(() => {
    filterMatchingOptionsRef(inputValue)
  }, [inputValue])
```

app/javascript/components/ReduxFormAutoComplete/index.js
ThuTrinh: I love ramda, the function composition features are awesome.

app/javascript/components/ReduxFormAutoComplete/index.js
ThuTrinh: actually @samzx , I did some research and it looks like the `useRef` is to prevent the debounce function from resetting between renders. Because if it did, there would be no debouncing.

a much more in-depth explanation here: https://overreacted.io/making-setinterval-declarative-with-react-hooks/

ThuTrinh (approved):

PR #275, opened 2019-12-31
Tags field for events and individual events

### LINKED ISSUE #265: Multiple event type fields.
It'd be nice to have generic fields to categories events. Currently we have one "Event Type" field. We should have multiple event type fields.

Have a title, admins can choose the field name ie: "Event Type" or "Focus Area"
and they both would share a similar structure where admins can add that type, users would need to select an option for these two. This is helpful for reporting in specific target areas like "Education & Local Neighbourhoods" or "Wellbeing & Families" or "Eduction & Families" etc.

samzx:
Closes #265 

Depends on #267 

Please be nice, I don't really know ruby 😅 

## Description
Adding Tags field for events and individual events. This helps admins have more detailed analysis in which combinations of event types users are attending.

In the future, we'll want to allow the ability for admins to control the number of tags users can add to to events, per tag group.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/265)

### Managing tags
image

### Adding tags to events (includes clone and edit)
image

### Adding tags to individual events (includes clone and edit)
image

## CCs

@zendesk/volunteer

## Risks
- medium


app/javascript/components/TagField/index.js
```diff
@@ -0,0 +1,95 @@
...
+      setMatchingOptions(matchingOptions)
+    }, 300)
+  )
+
+  useEffect(() => {
+    // Handles existing value and transforms to mutatable format
+    if (value) {
+      onChange(value.map(tag => ({ id: tag.id })))
+    }
+  }, [])
```
ThuTrinh:
```suggestion
  }, [value])
```

app/javascript/pages/admin/Tags/index.js
```diff
@@ -0,0 +1,183 @@
...
+import { graphQLError, togglePopover } from 'actions'
+
+import Loading from 'components/LoadingIcon'
+
+import TagsQuery from './queries/index.gql'
+import DeleteTagMutation from './mutations/delete.gql'
+
+import s from './main.css'
+
+import 'style-loader!css-loader!react-table/react-table.css'
```
ThuTrinh: I usually like to consistently organise my imports, it makes reading easier imo. We might need to come up with some standards and conventions for the portal going forward. But my convention here is to group and order them like so:

all react imports
newline
all third party imports
newline
all local imports

app/graphql/resolvers/tag_resolver.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def create(_, args, _context)
+      tag = Tag.new
+      tag.name = args[:input].name
+      tag.save!
+
+      tag
+    end
+
+    # def update(_, args, _context)
```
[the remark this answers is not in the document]
samzx: I think that should be at the admin's discretion. I think they should have the ability to rename say `Education` to `education` or swap "equivalences" around, but does raise a good point that Admins should be careful or not use this feature too frequently.

app/javascript/components/TagField/index.js
samzx: I thought of using `[]` so it loads the graphQL query (loaded once on component mount) into the redux form format. The `onChange` methods write to the redux form in the correct format so I guess it probably doesn't need to call this effect when the `value` changes.

app/graphql/resolvers/tag_resolver.rb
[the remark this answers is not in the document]
ThuTrinh: maybe it should include a confirm on the UI side.

app/javascript/pages/MyEvents/index.js
```diff
@@ -505,6 +505,7 @@ const individualEventInputToOptimisticResponse = (data, input) => {
     date: moment(input.date),
     duration: input.duration,
     eventType: R.find(et => et.id === input.eventTypeId, data.eventTypes),
+    tags: R.filter(tag => R.find(inputTag => inputTag.id === tag.id, input.tags), data.tags),
```
ThuTrinh: can also use `R.eqProps` the advantage here is it won't crash if `id` doesn't exist.
```
tags: R.filter(tag => R.includes(R.eqProps('id', tag), listB), listA)
```

ThuTrinh (approved):

ThuTrinh (approved):

Issue #164, opened 2019-06-05
Request for consistent UI - Garden Components


samzx:
A lot of the interfaces are not consistent. Sometimes using material ui, sometimes bare-bones components.

Might be a big task, but it would be nice to have all components derive from the Zendesk Garden components. https://garden.zendesk.com/

As referenced from: https://github.com/zendesk/volunteer_portal/pull/160#discussion_r286672557

samzx:
depends on #182

ThuTrinh:
### Dialog > Garden Modal
- /components/NamedAvatar.js
- /pages/Dashboard/index.js
- /pages/admin/EventTypes/index.js
- /pages/admin/Offices/index.js
- /pages/admin/Organizations/index.js
- /pages/admin/Tags/index.js
- /pages/admin/Users/index.js
- /pages/MyEvents/index.js

PR #301, opened 2020-01-17
Refactor Volunteer Leaderboard component to use Apollo hooks and Garden

ThuTrinh:
## Description
- Moved Leaderboard view and data fetch logic into its own component within the Dashboard folder
   - Refactored it to use Apollo hooks
   - Refactored it to use as much Garden components as possible
   - Added loading skeleton views for the user list, avatar and office filter
   - Add Garden style alert for error state
- Moved Milestones view and fetch logic into its own component within the Dashboard folder
   - Refactored it to use Apollo hooks
   - Removed dependency on Redux (by using `withRouter` to get current location)
- Updated the ShowFilter, EventFilter and OfficeFilter to use Garden dropdowns
   - Rearranged the Filters in places to fix alignment and spacing issues (refer to screen shots)
- NamedAvatar
   - Move the confirm delete modal logic out of NamedAvatar and into the UserList itself.
   - Removed dependency on redux
- UserList (Leaderboard & Event details)
   - Refactored to use as much Garden components as possible 
   - Updated the confirm delete modal to use Garden
- ListItem
   - Added a new ListItem component that is a box with a bottom border and when you place 2 things into it, they will be separated by a gap and hug the sides. 

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/164)

## Implementation notes
- Moving the Leaderboard and Milestones out into their own components has the following benefits:
   - Easier to compose and reason about
   - They are self contained and don't affect other components on the page when they fail, since they handle their own data fetching, error handling and loading states.
   - Easier to maintain. They are decoupled from everything else and can be easily moved around or deleted without much effort.
- Switched to using `styled-components` over css-loader.
   - Lets us keep the styles closer to where they are being used.
   - Easily see when styles are not being used during refactoring.  
   - We can leverage prop passing into our styles. This will become more useful when we enable themes properly.
   - This will be the standard for this project going forward.
- Each component should be responsible for its own loading skeleton. 
   - It should support a `loading` prop that determines whether the component should render its skeleton view instead. 
   - This keeps the skeleton loaders close to their real counterparts, enabling them to share any common layout styles. 
   - The `loading` prop gives us an easy interface to trickle down loading state and compose skeleton loaders.
- I chose not to refactor the styling of Milestones too much as I think this component needs a redesign anyway. As it stands, I don't think it is very useful.

## Screenshots
<details>
<summary>Before</summary>
<strong>Loaded State</strong>
<hr/>

<strong>Loading State</strong>
<hr/>

<strong>Misaligned Filter
</strong>
<hr/>

</details>

<details>
<summary>After</summary>
<strong>Loaded State</strong>
<hr/>

<strong>Loading State</strong>
<hr/>

<strong>Error State</strong>
<hr/>

<strong>Aligned/Repositioned Filter</strong>
<hr/>

<strong>Aligned/Repositioned Filter</strong>
<hr/>

<strong>User List
</strong>
<hr/>

</details>

## CCs
@zendesk/volunteer

## Risks
* [medium] - there might be some regression because some common components are changed. A full manual test session will need to be done on stg.


Pancrisp:
Yessss, skeleton loaders! I was just thinking of swapping out the loaders we currently have for the skeleton ones.

Thanks for kicking things off and getting the ball rolling.

app/javascript/components/OfficeFilter/index.js
```diff
@@ -1,22 +1,53 @@
...
-import { styles } from 'components/Filter'
+const Loader = withNamespaces()(({ t }) => (
+  <Dropdown>
+    <Field>
+      <Select selectedItem="">
+        <strong>{t('volunteer_portal.dashboard.layouteventstab.office')}</strong> <Skeleton width="4rem" />
+      </Select>
+    </Field>
+    <Menu>
+      {[1, 2, 3, 4].map(i => (
```
samzx: What's your opinion on `Array(4).fill().map((_, i) => ...` instead? I think you used it that way in `app/javascript/pages/Dashboard/Leaderboard/index.js`

app/javascript/components/Toolbar/index.js
```diff
@@ -39,12 +39,16 @@ const Toolbar = ({ label, view, views, onNavigate, onViewChange, filters, office
         </button>
       </div>
       <div className={s.filterBar}>
-        <div className={s.filterDropdowns}>
-          <ShowFilter value={filters.showFilter.value} onChange={filters.showFilter.onChange} />
-          <EventFilter value={filters.eventFilter.value} onChange={filters.eventFilter.onChange} />
-          <OfficeFilter value={filters.officeFilter.value} onChange={filters.officeFilter.onChange} offices={offices} />
+        <div className="u-display-inline-flex u-mb-xxs">
```
samzx: I wonder if it's bad to be mixing the use of both the spacing utility classNames with styled components. I got kind of confused where these class names came from (but then I googled ha).

Is it possible to extract the spacing utilities from https://garden.zendesk.com/css-components/utilities/spacing/ and access them via styled components and use `styled.div` or something along that line?

Maybe not in this PR, but just a thought for the future 🤔

app/javascript/components/UserList/index.js
```diff
@@ -1,26 +1,79 @@
...
-            showRemove={present(destroySignup)}
-            onRemove={() => destroySignup(user)}
-          />
-        </div>
-      ))}
-  </div>
-)
+      {activeUser && (
+        <Modal onClose={cancelRemove}>
+          <Header>🤔Remove volunteer from this event? </Header>
```
samzx: love this 😹

app/javascript/pages/Dashboard/Leaderboard/index.js
```diff
@@ -0,0 +1,107 @@
...
+    variables: {
+      after: startOfYear,
+      before: nowInSec,
+      count: leaderBoardSize,
+      officeId: currenOfficeId,
+      sortBy: leaderBoardSort,
+    },
+  })
+
+  const ofifces = R.propOr([], 'offices', data)
```
samzx: offices? 😂

app/javascript/pages/Dashboard/Leaderboard/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+// The server expects seconds since epoch, not milliseconds
+const startOfYear = Math.floor(moment().startOf('year') / 1000)
+const nowInSec = moment().unix()
+const leaderBoardSize = 10
+const leaderBoardSort = 'HOURS_DESC'
+
+const Leaderboard = _props => {
+  const currentUser = useContext(UserContext)
+  const [currenOfficeId, setCurrentOfficeId] = useState(R.path(['office', 'Id'], currentUser))
```
samzx: currentOfficeId 😛

app/javascript/pages/Dashboard/Leaderboard/index.js
```diff
@@ -0,0 +1,107 @@
...
+  padding-bottom: ${zdSpacingXxs};
+`
+const SectionTitle = styled(LG)`
+  align-self: flex-end;
+`
+
+const Loader = _props => (
+  <>
+    {Array(10)
+      .fill(null)
```
samzx: is the `null` necessary? - I think it still works if you map over undefined?

samzx (commented):
Awesome work 🎉 this is really, really great, I'm so happy you did this! 😊

app/javascript/components/OfficeFilter/index.js
ThuTrinh: I didn't here because 4 is such a small number and I save on an iteration. But I am not really fused. If it's for consistency then I can change it.

app/javascript/components/Toolbar/index.js
ThuTrinh: I also prefer styled components over the class names. I can access the spacing values through the garden css variables. I will do that instead :)

app/javascript/pages/Dashboard/Leaderboard/index.js
ThuTrinh: woops

app/javascript/pages/Dashboard/Leaderboard/index.js
ThuTrinh: yeah it should still work.

app/javascript/components/Toolbar/index.js
ThuTrinh: @samzx what are your thought on using something like this? https://styled-system.com
I have used it before and really like it. When done right, you end up with far less components.

samzx (approved):
🙌 Very nice ! Think you just [missed](https://github.com/zendesk/volunteer_portal/pull/301/files#diff-14bf74813f38072a8c0ed2bed9e3524bR65) `ofifces`, otherwise LGTM

app/javascript/components/Toolbar/index.js
samzx: Looks like an interesting concept 😮 I'll have to have a try to see what it is

PR #303, opened 2020-01-19
Upgrade Node to latest LTS - v12

Pancrisp:
## Description
We're still using Node 8 which reached EOL last month.

Thankfully, Node versions 8 and up aren't too different from one another, so they're _mostly_ backwards compatible.

From my initial checks, the upgrade did not break any of our dependencies.
We don't have many tests for our front-end app, but I did some manual testing and didn't see anything unusual.

## CCs
@zendesk/volunteer

## Risks (if any)
* Medium - Some components might stop working after the upgrade


ThuTrinh (approved):
nice one

PR #307, opened 2020-01-29
Update translation plugin

samzx:
## Description
We've decided we don't want to tie ourselves to Rosetta from Zendesk, so we're altering the translation pipeline.

How this one works is that it will transform csv files into a serve-able json file. The csv files should be synced with each other and translations of the csv files. Generally we'd be extracting the strings into the en.csv file and translating a es.csv ja.csv etc. files for other languages.

## CCs

If app migration related
@zendesk/volunteer 

## Risks (if any)
* low


config/locales/README.md
```diff
@@ -0,0 +1,36 @@
...
+Add translations in `en.csv` file in the following format.
+
+```
+key,value,title,screenshot,wip
+volunteer_portal.dashboard.personal,Label referring to who the number of hours belong to,Personal,,true
+volunteer_portal.dashboard.layouttab.calendar,Clickable tab that displays the Calendar View,Calendar,https://drive.google.com/open?id=1TEM84-1i6-GgF7N4cJuoTouGgCi0HC_f,true
+
+```
+
+Our preffered method is using the `useTranslation` hook, but there's a lot of legacy using the `withTranslation` after converting from v9.
```
ThuTrinh:
```suggestion
Our preferred method is using the `useTranslation` hook, but there's a lot of legacy using the `withTranslation` after converting from v9.
```

ThuTrinh (approved):

PR #310, opened 2020-01-29
Ignore unapproved individual event hours

kennethc:
Ignore unapproved individual event hours in Dashboard Top Volunteers

## Description
The new VolunteerResolver did not take into account the approval status of individual events. This caused the Dashboard to show volunteers with many unapproved individual event hours, which manifests as an unsorted dashboard.

This change will make VolunteerResolver count only the approved individual events as it should be.

## References
N/A

## CCs

If app migration related
@zendesk/volunteer @zendesk/mel-volunteer 

## Risks (if any)
* low - no impact expected but could have unforeseen risks


samzx (approved):
👏 nice work :)

Issue #311, opened 2020-02-05
Only admins should be able to remove users from an event


ThuTrinh:
It would appear that a regression bug made it's way through a recent [refactor](https://github.com/zendesk/volunteer_portal/commit/d7926badcfd4d4d87444a4aad653649a27129812#diff-e0cb136fdc01a84b2bdff835ea25f23e).

This bug is localised to the front end only and doesn't appear to be causing too much harm. The remove bin icon appears regardless of what role the current user has. However when actioned, the request is denied on the server causing an error that can only be seen in the browser console. 

<img width="1997" alt="Image 2020-02-03 at 2 39 07 pm" src="https://user-images.githubusercontent.com/3032977/73835941-69f6d380-4862-11ea-9174-52cfcbedd522.png">

PR #312, opened 2020-02-05
Only admins can remove users from an event

ThuTrinh:
## Description
When displaying the list of users that have signed up for an event, only show the remove button if the current signed in user is an admin.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/311)

## Screenshots
vp-admin-only-delete-after

## Implementation Details
The fix for this issue is only relevant in 1 file: `app/javascript/pages/Event/index.js`

I decided to just refactor the components a bit to make it clearer and more maintainable.

## CCs
@zendesk/volunteer

## Risks (if any)
* [low] - minor change to hide a button, should not effect anything else.


app/javascript/components/UserList/index.js
```diff
@@ -45,14 +47,16 @@ const UserList = ({ users, destroySignup }) => {
...
-              <IconButton
-                aria-label={`Remove volunteer ${user.name} from this event`}
-                onClick={() => setActiveUser(user)}
-              >
-                <Icon>
-                  <RemoveIcon />
-                </Icon>
-              </IconButton>
+              {currentUser.isAdmin && (
+                <IconButton
```
kennethc: This is technically correct, but is it possible to use an `if` block here? I think it will be clearer to a wider audience what we are trying to achieve here.

ThuTrinh: sure :)

ThuTrinh: I gave this more thought and while I do agree with you, I think that we should keep the statement as is for the following reasons:
1. This technique of rendering logic is idiomatic in react and will be present in many codebases. We should expose people to it and help them use it appropriately rather than avoid it imo.
2. The react [documentation](from './UserList') encourages it anyway, with a caveat that the value before the `&&` must be a boolean. We should look for this in code reviews.
3. The alternative `if` statement actually makes the code look more complicated imo. 
e.g.
```javascript
// inline
{currentUser.isAdmin ? (
    <IconButton aria-label={`Remove volunteer ${user.name} from this event`} onClick={confirm}>
        <Icon>
          <RemoveIcon />
        </Icon>
      </IconButton>
    ) : null 
}
```

Pancrisp (approved):
Agree with @ThuTrinh's statement above. The logical `&&` operator is a standard approach to conditional rendering in React and is preferred over the ternary operator if your false conditional doesn't output anything.

Issue #178, opened 2019-07-31
Regularly update user information (ie. departments)


samzx:
Currently the user departments are old. It should be updated on a regular basis, either from google or okta. This is also useful for keeping reporting and analytics up to date.

Perhaps weekly cron that updates all user data to our database.

yoshdog:
API to get it from google:
https://developers.google.com/admin-sdk/directory/v1/reference/users/list#auth

API to get it from Okta:
https://developer.okta.com/docs/reference/api/users/#list-users

Will test it out for both and see whats easier.

PR #313, opened 2020-02-12
remove subtitle for departments

samzx:
Removing departments until we have https://github.com/zendesk/volunteer_portal/issues/178 solved, we have users mentioning that it's not correct. We'd rather not show the data unless it's correct.

cc @ThuTrinh  @azaiszen

### DIFF
--- a/app/javascript/components/NamedAvatar.js
+++ b/app/javascript/components/NamedAvatar.js
@@ -40,7 +40,7 @@ const NamedAvatar = ({ loading, image, name, subtitle }) => {
       <Avatar image={image} />
       <Details>
         <MD tag="strong">{name}</MD>
-        <SM>{subtitle || '\u00a0'}</SM>
+        {subtitle && <SM>{subtitle}</SM>}
       </Details>
     </Container>
   )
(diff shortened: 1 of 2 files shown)


app/javascript/components/NamedAvatar.js
```diff
@@ -33,14 +33,13 @@ const Loader = _props => (
...
 
-const NamedAvatar = ({ loading, image, name, subtitle }) => {
+const NamedAvatar = ({ loading, image, name }) => {
   if (loading) return <Loader />
   return (
     <Container>
       <Avatar image={image} />
       <Details>
         <MD tag="strong">{name}</MD>
-        <SM>{subtitle || '\u00a0'}</SM>
```
ThuTrinh: instead of removing the functionality entirely, can we just not pass it in?

samzx: Yep, that'll be easier to hook up when we do get the departments to sync, I just went on a cleaning spree 😄 i'll change this

ThuTrinh (approved):
Looks good apart from the 1 comment

PR #318, opened 2020-02-13
Remove debugging console.log

kayison:
## Description
This console.log was accidentally left in from debugging, so it is removed.

## Risks (if any)
*  Low: only console.log removal


anita-miller (approved):
💯

ThuTrinh (approved):

Issue #319, opened 2020-02-14
Remove/Wire-up Office Filter in Admin Pages


ThuTrinh:
All pages in the admin section currently displays the office filter, even if it doesn't apply or is not wired up. This is confusing to the user.

### Task
- Update the admin pages so that each tab is responsible for displaying their own filter
   - Each page can then have different filters or no filters, what ever is applicable
- Update the graph queries for users and individual events to support the office filter

PR #321, opened 2020-02-16
admin section: each tab should provide their own filters

ThuTrinh:
## Description
This PR is in regards to the office filter in the Admin section. Currently the filter is provided at a parent level and shared by the contents of each tab.
Unfortunately this approach implies that the office filter applies to every tab content, which it doesn't.

This PR is to move the responsibility of providing these filters to the tab content components. This means that the filter will not be present where they are not applicable.

Also note although the office filter can apply to the individual events and users tabs, they were never wired up. I have removed them in this PR and address wiring them up to the office filter in a separate PR.

### Changes
- updated date picker in the report section to use garden date picker
- updated action links (add *) to garden buttons

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/319)

## Screenshots
<details>
<summary>Before</summary>

</details>

<details>
<summary>After</summary>

</details>

## CCs
@zendesk/volunteer

## Risks
* [medium] - Report filter needs testing on staging, just to make sure the date picker changes work as expected.


app/javascript/pages/admin/Reporting/index.js
```diff
@@ -9,12 +9,13 @@ import Reporting from 'components/Reporting'
...
 import ReportingQuery from './query.gql'
 import s from './main.css'
+import { addDays, compareAsc } from 'date-fns'
 
 const defaultStartDate = moment().startOf('year')
 const defaultEndDate = moment().valueOf() // Now in Unix millisecond timestamp
 const initialRange = {
-  start: undefined,
-  end: undefined,
+  start: addDays(new Date(), -14),
```
samzx: nit: extract out magic number as maybe fornight?

app/javascript/components/Reporting/index.js
```diff
@@ -105,35 +95,39 @@ const tableExporter = (startDate, endDate, onStartChange, onEndChange, state, ma
   )
 
   const octetStream = encodeURIComponent(csv)
+  const today = new Date()
```
samzx: Do we want the date being updated every re-render? Or just per page load - then we can call it outside the function / useEffect (for once only).

ThuTrinh: well, its kind of inexpensive and there is the edge case where the component mounts before 12am and reloads after 12am 🤷‍♀

samzx (approved):

Issue #107, opened 2018-11-19
Add Support for Google Tag Manager


rshokrizadeh:
Rather than supporting individual providers, we should add support for the developer to specify the container ID as a variable and it gets properly added to the snippet in the code base. This will allow people to add other tags into the app without needing to edit the code base directly.

**High in head:**

<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','INSERT CONTAINER ID HERE');</script>

**After the opening <body> tag:**

<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=INSERT CONTAINER ID HERE"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>

Will need to update the docs to explain where to specify the variable, similar to what we do currently for the Web Widget implementation.

azaiszen:
![](https://github.trello.services/images/mini-trello-icon.png) [Tagging events - Adding a volunteering type. Physical Volunteering vs Planning or Meetings etc.](https://trello.com/c/9JHmVHMK/6-tagging-events-adding-a-volunteering-type-physical-volunteering-vs-planning-or-meetings-etc)

samzx:
resolved by #323

rdzudzar:
Hi, I have a question about positioning Zendesk chat via GTM.

I have Zendesk Chat Widget GTM pixel code on clients website. And I found online that one can change position with the code below (from [developer zendesk](https://developer.zendesk.com/api-reference/widget/settings/?_ga=2.165456831.1600603858.1638319047-1290681618.1638319047#position)).
However, When I add custom html tag with this code - Zendesk chat does not change position at all. 
In addition, on my Zendesk dashboard: widget appearance has only option bottom right/left.

```
<script type="text/javascript">
  window.zESettings = {
    webWidget: {
      position: { horizontal: 'left', vertical: 'top' }
    }
  };
</script>

```

Thanks,
Robert

PR #323, opened 2020-02-26
add google tag manager

samzx:
solves #107  - allows us to add analytics and other static html to help analyse the site.

seems to work in staging! :)

### DIFF
--- a/app/javascript/index.js
+++ b/app/javascript/index.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
+import TagManager from 'react-gtm-module'
 import { ApolloClient } from 'apollo-client'
 import { HttpLink } from 'apollo-link-http'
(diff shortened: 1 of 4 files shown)


app/javascript/index.js
```diff
@@ -14,6 +15,12 @@ import { lightTheme } from '/themes'
 import './i18n'
 import './styles'
 
+const tagManagerArgs = {
+  gtmId: process.env.GOOGLE_TAG_MANAGER_ID,
+}
+
+TagManager.initialize(tagManagerArgs)
```
ThuTrinh: `GOOGLE_TAG_MANAGER_ID` is marked as optional, what would happen when it is not set?

samzx: good point, I'll initialise it only if it's set

(it will send a request and get a 404 if it's not set)

ThuTrinh (approved):

PR #324, opened 2020-02-26
admin can add volunteers manually on Event edit page

liulikun:
## Description
Add a new feature that allows admin to add individual volunteer by the email on Event edit page.

Need to disable `Add` button when the email address is blank/invalid.

## Tasks (if needed)
- [ ] Write tests

## Screenshots (if needed)
<details>
<summary>Before</summary>
Screen Shot 2020-02-26 at 3 38 45 pm

</details>

<details>
<summary>After</summary>
Screen Shot 2020-02-26 at 3 36 54 pm

</details>

## CCs

If app migration related
@zendesk/volunteer

Anyone specific?

## Risks (if any)
* [HIGH | medium | low] - low

### DIFF
--- a/app/graphql/mutations/create_signup.rb
+++ b/app/graphql/mutations/create_signup.rb
@@ -5,10 +5,12 @@ module Mutations
     null true
 
     argument :event_id, ID, required: true
+    argument :user_email, String, required: false
 
-    def resolve(event_id:)
+    def resolve(event_id:, user_email: nil)
       event = Event.find(event_id)
-      event.sign_up_user!(context[:current_user])
+      user = user_email.present? ? User.where(email: user_email).first : context[:current_user]
+      event.sign_up_user!(user) if user
       event.save!
       event.signups.last
     end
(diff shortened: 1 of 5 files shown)


app/graphql/mutations/create_signup.rb
```diff
@@ -5,10 +5,12 @@ class CreateSignup < BaseMutation
     null true
 
     argument :event_id, ID, required: true
+    argument :user_email, String, required: false
 
-    def resolve(event_id:)
+    def resolve(event_id:, user_email: nil)
```
ThuTrinh: This resolver was originally designed to leverage the current session user. Since the current user would always have permission to sign themselves up for an event, it was safe.

However, adding an optional email argument without any authorisation, would mean that any user would be able to sign any other user up for an event.

Alternatively, separating the mutations would make more sense here imo:
- They are different operations and require different inputs. i.e. email is not actually optional in the context of an admin adding a person to an event.
- They require different permissions/roles. i.e. only admins should be able to access this mutation.
- The clear separation will help maintainability and dev onboarding

PR #337, opened 2020-04-24
Garden components for "Add individual Event"

### LINKED ISSUE #156: As a user (volunteer and admin), I want creating events to be easy
Current forms are clunky and time intensive to use. Users dread making events because of the bad user experience with the forms.

There should be a new version of forms which is **intuitive, quick and simple**

samzx:
Closes #156 

Update "adding new events" (individual events) to use garden components

# Before
image

image

# After
image

image

## Description
- Why single column? Eye to flow top down as opposed to LTR, then down. Since we're not using many fields, we don't need multi-column to make the form seem smaller. Different field types make it hard to accommodate a similar height for each field, making it look awkward.

- Ordering has been changed to be more logical. Humans tend to recall location and time first. Then we have the user fill in details around that event. Finally, some extra meta-data which isn't really related to the event, but the office you belong to. (in-out flow)

- Duration has a slider from 0-8 hours. This is much less confusing than the 'minutes' where the user has to convert to minutes. We cap at 8 hours since, most likely, a user won't volunteer longer than a typical working day. If they do, they should record two events. Extra usability, but sacrifices these edge cases.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/XX)

## Implementation notes
- Everything related to individual events have been refactored to use Garden components. The buttons to activate the modal too.

## Risks
* low


app/javascript/pages/MyEvents/CreateEditModalContents.jsx
```diff
@@ -0,0 +1,174 @@
...
+  float: right;
+`
+
+const CreateEditModalContents = ({
+  offices,
+  eventTypes,
+  tags,
+  organizations,
+  createEditIndividualEvent,
+  isNew,
```
ThuTrinh, line 32: I feel like if we are going to have this property, we should validate that `modalEventData` is not null and has a nonzero id when `isNew=false`

Because looking at how the data is submitted to the server, a new event will be created in the absence of an id.

app/javascript/pages/MyEvents/CreateEditModalContents.jsx
```diff
@@ -0,0 +1,174 @@
...
+        <PaddedField>
+          <Label>Description</Label>
+          <Input value={description} onChange={event => setDescription(event.target.value)} />
+          {showFieldErrors && !isValidText(description) && (
+            <Message validation={'error'}>{'Must not be empty'}</Message>
+          )}
+        </PaddedField>
+        <PaddedField>
+          <Label>Organization</Label>
+          <Hint>Contact us to add your organization to this list or select 'Other'</Hint>
```
ThuTrinh, line 96: How do they contact us?

app/javascript/pages/MyEvents/IndividualEvents.jsx
```diff
@@ -0,0 +1,237 @@
...
+import CreateEditModalContents from './CreateEditModalContents'
+import DeleteModalContents from './DeleteModalContents'
+import { NoEventsMessage, EventsTable } from './StyledComponents'
+
+import styled from 'styled-components'
+
+const eventStatusIcon = event => {
+  switch (event.status) {
+    case 'REJECTED':
+      return <AVNotInterested color="#A93A38" />
```
ThuTrinh, line 22: Should we use garden color variables here?

app/javascript/pages/MyEvents/IndividualEvents.jsx
```diff
@@ -0,0 +1,237 @@
...
+  max-width: 200px;
+  white-space: nowrap;
+`
+
+const PersonalHeader = styled.div`
+  position: relative;
+`
+
+const IndividualEvents = props => {
+  const { data, popover, togglePopover, handleSubmit, createEditIndividualEvent, deleteIndividualEvent } = props
```
ThuTrinh, line 56: Looks like `togglePopover` and `handleSubmit` are unused

app/javascript/pages/MyEvents/DeleteModalContents.jsx
```diff
@@ -0,0 +1,40 @@
...
+    setLoading(true)
+    deleteIndividualEvent(toDelete.id).then(() => {
+      setLoading(false)
+      setShowDeleteModal(false)
+    })
+  }
+
+  return (
+    <>
+      <Header isDanger>Delete Event</Header>
```
ThuTrinh, line 19: Should this be danger? usually we use danger to represent an error. This is more of a warning imo.

ThuTrinh (approved):
Just some minor suggestions/questions.
The rest LGTM.

ThuTrinh:
would setting a max limit of hours to 24 make more sense than 8?

samzx:
> would setting a max limit of hours to 24 make more sense than 8?

I wasn't sure about this but in most cases a user would volunteer less than a working day, and if they did more they'd record twice? Slight sacrifice for some better usability?

app/javascript/pages/MyEvents/DeleteModalContents.jsx
samzx: I think danger is fine to represent a dangerous action (such as deletion)

app/javascript/pages/MyEvents/IndividualEvents.jsx
samzx: I'll see if we can do that :)

app/javascript/pages/MyEvents/CreateEditModalContents.jsx
samzx: image

We've got the support widget. Not everyone will have it but I suspect there'll be a point of contact (the admins) in which they can add organisation. This will be how it is for now until volunteers can add organizations.

app/javascript/pages/MyEvents/DeleteModalContents.jsx
ThuTrinh: true

ThuTrinh:
> if they did more they'd record twice? Slight sacrifice for some better usability?

It is up to you, I am not fussed either way. I just thought 24 would be more versatile because it is the actual limit of a day.

app/javascript/pages/MyEvents/IndividualEvents.jsx
samzx: changed the icons to garden icons 😃

PR #338, opened 2020-05-06
Populate tags for volunteer portals without tags.

samzx:
## Description
This is because all events require at least one tag, for now. In the future when there's an option to make it optional, we can remove this migration.

This is especially important for old volunteer portals that have migrated on to with tags. Not so important for new portals since they have the seed file.

## Tasks (if needed)
- [x] Tested locally

## CCs
If app migration related
@zendesk/volunteer

## Risks (if any)
* [HIGH | medium | low] - low

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20200114001524) do
+ActiveRecord::Schema.define(version: 20200506054322) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -60,11 +60,6 @@
     t.index ["updated_at"], name: "index_events_on_updated_at"
   end
 
-  create_table "events_tags", id: false, force: :cascade do |t|
-    t.bigint "event_id", null: false
-    t.bigint "tag_id", null: false
-  end
```
samzx, line 66: think this was auto-generated from `rails g migration no_empty_tags description:string` not sure why though. This table isn't used anywhere in the past anyway. we use `event_tags` for event-tag relationship.

ThuTrinh (approved):

PR #341, opened 2020-05-08
Allow gauth login from clients (non-rails)

samzx:
## Description
This exposes `/gauth` endpoint that allows clients such as ios, android, web clients (non server side) to exchange a google auth token for a session. It only authorises users, it does not support user creation yet. Once a user is authorised, they are able to use graphql queries.

Note: for a user to be authorised, they'll need to sign in with google.

## References
https://github.com/zendesk/volunteer_portal/issues/340

## CCs
@zendesk/volunteer

## Risks (if any)
* medium


samzx:
closed in favour of https://github.com/zendesk/volunteer_portal/pull/380

PR #349, opened 2020-05-20
Upgrade gems

kayison:
## Description

Upgrade gems in preparation for rails upgrade. Some of these are major version bumps and we might want to do these separately if we think they're too high of a risk.

## Risks (if any)
* Medium - there are some major version bumps in packages that should be tested in staging


samzx (approved):

kayison:
Dependabot PRs have this up to date

PR #350, opened 2020-05-20
Upgrade rails 5.1.6.2 to 5.2.4.3

kayison:
## Description

Upgrade rails to fix security vulnerabilities in earlier versions. 

There are some Rails 6 deprecation errors that have been silenced temporarily to enable us to upgrade to fix the security vulnerabilities. 

Some rubocop changes as part of the upgrade. Add ignores to existing migration files so we don't modify them but leave the rule open for new ones.

## CCs

If app migration related
@zendesk/volunteer

Anyone specific?

## Risks (if any)
* Low: test to make sure rails upgrade is ok


db/migrate/20161211042811_add_timestamps_to_offices.rb
```diff
@@ -1,7 +1,9 @@
+# rubocop:disable Rails/BulkChangeTable
 class AddTimestampsToOffices < ActiveRecord::Migration[4.2]
   def change
     add_column :offices, :created_at, :datetime
```
samzx, line 4: not too familiar with rails, but it seems like these can be combined into one somehow?

samzx (approved):
we can have a look at combining the queries in another pr if possible

db/migrate/20161211042811_add_timestamps_to_offices.rb
kayison: Do you mean adding something to the `.rubocop.yml`?

samzx: never mind, I think this is fine, unless we can ignore the directory somehow

kayison: I'm ignoring existing migrations so we don't modify them and keeping the rule open for any new migrations

PR #355, opened 2020-05-27
fix translation keys

samzx:
Some of the keys were using `.` instead of `_` which didn't show the correct keys. This fixes that.


ovillegaszen (approved):
👌

PR #381, opened 2020-06-19
Update vendor/cache after dependabot creates gem PR

kayison:
Update vendor/cache directory after dependabot creates a PR for gems

## Description
- Dependabot does not yet update `vendor/cache` for us. This GitHub Action will run when a dependabot PR is created for a Ruby gem and add `vendor/cache`.
- Dependabot generates PRs for us when packages need upgrading or there are security vulnerabilities. However there is no ability (at least not yet) to auto update our vendor/cache files. This means we must check out each branch created by dependabot and run bundle and commit manually to get the vendor/cache files.

Please note that this action will stop further rebases of the branch to either rebase of master or to rebase when a newer version is available.

## CCs

If app migration related
@zendesk/volunteer

Anyone specific?

## Risks (if any)
* Low: build should break if this doesn't work

### DIFF
--- a/.github/workflows/update-gems-after-dependabot.yml
+++ b/.github/workflows/update-gems-after-dependabot.yml
@@ -1,0 +1,40 @@
+---
+name: Update Gems after Dependabot
+
+on:
+  push:
+    branches:
+      - 'dependabot/bundler/**'
+    paths:
+      - 'Gemfile.lock'
+
+jobs:
+  update-gems:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Ruby and Bundler
+        uses: actions/setup-ruby@v1
+
+      - name: Cache gems
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: bundler-${{ runner.os }}-${{ hashFiles('.ruby-version') }}-${{ hashFiles('Gemfile.lock') }}
+          restore-keys: |
+            bundler-${{ runner.os }}-${{ hashFiles('.ruby-version') }}-
+
+      - name: Install gems
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+
+      - name: Push a commit with changes to vendor/cache
+        run: |
+          git config --local user.email `git show -s --format='%ae'`
+          git config --local user.name ${GITHUB_ACTOR}
+          git remote add github "https://$GITHUB_ACTOR:$GITHUB_TOKEN@github.com/$GITHUB_REPOSITORY.git"
+          git add vendor/cache
+          git commit -m "`git log --oneline --format=%B -n 1 HEAD | head -n 1` - add gems"
+          git push github HEAD:${GITHUB_REF}


samzx (commented):
Wondering if this can be run on the standard runners since this is open source? Or we might be limited because this project is under the Zendesk org so we have limited hours?

Similarly wondering if there's something else with free open source usages instead of jfrog that's visible to non-zendesk employees.

.github/workflows/update-gems-after-dependabot.yml
```diff
@@ -0,0 +1,43 @@
...
+  on:
+    push:
+      branches:
+        - 'dependabot/bundler/**'
+      paths:
+        - 'Gemfile.lock'
+  
+  jobs:
+    update-gems:
+      runs-on: zendesk-stable
```
craig-day, line 13: This should be `runs-on: ubuntu-latest` since this project is OS

.github/workflows/update-gems-after-dependabot.yml
```diff
@@ -0,0 +1,43 @@
+---
+  name: Update Gems after Dependabot
```
craig-day, line 2: this extra level of indentation isn't necessary in YAML

.github/workflows/update-gems-after-dependabot.yml
```diff
@@ -0,0 +1,43 @@
...
+      env:
+        BUNDLE_ZDREPO__JFROG__IO: "${{ secrets.BUNDLE_ZDREPO__JFROG__IO }}"
+      steps:
+        - uses: zendesk/checkout@v2
+  
+        - name: Install Ruby and Bundler
+          uses: zendesk/setup-ruby@v1
+  
+        - name: Cache gems
+          uses: zendesk/cache@v1
```
craig-day, line 23: `v2` of cache is out

craig-day (dismissed):
Looks good except we don't need our own self-hosted runners here

.github/workflows/update-gems-after-dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+    branches:
+      - 'dependabot/bundler/**'
+    paths:
+      - 'Gemfile.lock'
+
+jobs:
+  update-gems:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: zendesk/checkout@v2
```
samzx, line 15: can we use the OS actions/checkout@v2 here?

.github/workflows/update-gems-after-dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Install Ruby and Bundler
+        uses: zendesk/setup-ruby@v1
```
samzx, line 18: similarly ruby/setup-ruby@v1

.github/workflows/update-gems-after-dependabot.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Cache gems
+        uses: zendesk/cache@v2
```
samzx, line 21: and actions/cache@v2

samzx (commented):
looks good! Just wondering if we can use the OS actions instead ie.  https://github.com/actions/checkout

samzx (approved):
lgtm

Issue #226, opened 2019-10-21
Redirect to destination after log-in


alanhogan:
I clicked a portal link, but after signing in with SSO, I was sent to the home page instead of the URL of the link I clicked on. This URL should be propagated through the sign-in process.

samzx:
resolved by #407

Let me know if there are issues with this 🙂

PR #407, opened 2020-07-15
Save request path before redirecting to login screen. resume url after authentication

samzx:
Solves #226 

## Description
Say a user wants to visit an event `http://localhost:5000/portal/events/81`, if they're not already authenticated, it'll send them to the login screen. Then after login they' will be redirected. Before this PR, they'll always be sent to the home screen. After this PR, they will be sent back to the original url.

## Risks
* low


anita-miller (approved):

samzx:
resolves #226

PR #428, opened 2020-09-11
Refactor reporting section to allow multiple reports

samzx:
## Description
- Previously, the reporting section was tied to just user reports. This refactor allows extension of the reporting section to allow reporting of different dimensions ie. tags, event types, organizations.
- The date picker is no longer tied to the table - it will be available for all reports.
- The export csv is no longer coupled to the table. The octect stream is now dynamic depending on the report. It sources directly from the graphql query so we don't need to always display the full table before exporting.
- The office filter is now table specific, since it's only relevant to the report (users). Though we could have the office filter at the top level too.
- along with the refactor, I've moved to use styled components, react garden components and renamed to jsx file.

Missing features
- table columns can no longer be resized (resulting from moving to garden tables)
- tables can no longer be sorted by hours. I'll do this at a later date, along with pagination. Not a huge deal since admins can export and sort in excel or something.

## References
[Github Issue](https://github.com/zendesk/volunteer_portal/issues/XX)

## Screenshots (if needed)
<details>
<summary>Before</summary>
Screen Shot 2020-09-11 at 10 59 29 am
</details>

<details>
<summary>After</summary>
Screen Shot 2020-09-11 at 12 35 18 pm
</details>

## CCs

If app migration related
@zendesk/volunteer

## Risks
- Low


app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,254 @@
...
+
+  const handleDropdownOnSelect = item => {
+    if (item !== 'general' && item !== 'individual'  && item !== 'organized') {
+      setReport(item)
+    }
+  }
+
+  const renderMenu = () => (
+    <Menu >
+      {dropDownTempSelectedItem === 'organized' ? (
```
ThuTrinh, line 155: This is fine as is, but an alternative pattern that is easier to read and maintain would be to split each case into it's own component and use ramda's `R.cond` ([docs](https://ramdajs.com/docs/#cond)) to switch based on the value of `dropDownTempSelectedItem`.

Typically we only have a component per file so subcomponents usually live in the same folder or a `pages/admin/Reporting/components` folder. 

e.g.
```
      {R.cond([
        [R.equals('organized'), R.always(<OrganisedMenu />)],
        [R.equals('individual'), R.always(<IndividualMenu />)],
        [R.T, R.always(<MainMenu />)],
      ])(dropDownTempSelectedItem)
```

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,254 @@
...
+          <Item value="individualEvent-type">{t('volunteer_portal.admin.tab.reporting.dropdown.nested.event_type')}</Item>
+          <Item value="individualTag">{t('volunteer_portal.admin.tab.reporting.dropdown.nested.tag')}</Item>
+          <Item value="individualOrganization">{t('volunteer_portal.admin.tab.reporting.dropdown.nested.organization')}</Item>
+          <Item value="individualTotal">{t('volunteer_portal.admin.tab.reporting.dropdown.nested.total')}</Item>
+        </>
+      ) : (
+        <>
+          <Item value="user">{t('volunteer_portal.admin.tab.reporting.dropdown.users')}</Item>
+          <NextItem value="organized" disabled>{t('volunteer_portal.admin.tab.reporting.dropdown.organized_events')}</NextItem>
+          <NextItem value="individual" disabled>{t('volunteer_portal.admin.tab.reporting.dropdown.individual_events')}</NextItem>
```
ThuTrinh, line 177: I don't know about this tbh.
1. From an end user perspective, It might feel like a bug. I don't think this communicates that it is an unimplemented feature.
2. I don't think unimplemented features have any business being in the UI. Usually disabled options means that it will be enabled when certain conditions are met (i.e. permissions or conditional data). 
3. There is a lot of complex switching logic to accomodate this unimplemented. This logic will run unnecessarily, adds surface area for bugs and makes it hard to reason about imo.

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,254 @@
...
+            >
+            <Field style={{ width: 200 }}>
+              <FormLabel>{t('volunteer_portal.admin.tab.reporting.select_label')}</FormLabel>
+              <Select>
+                {
+                  report === null ? t('volunteer_portal.admin.tab.reporting.dropdown.none') : reportKeyTitleMap[report]
+                }
+              </Select>
+            </Field>
+            {renderMenu()}
```
ThuTrinh, line 200: This is fine, but fyi typically we wouldn't call a local function to render a component like this. Normally the Menu code would be in its own component and we would import it in. The reasoning is that it is easier to see the separations of concerns that way which helps with readability and maintenance.

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,254 @@
...
+          </a>
+        </ToolbarHeader>
+        {
+          report === 'user' ?
+          (
+            userLoading ?
+            <div>
+              <Skeleton height="60px" />
+              {
+                Array(10).fill(0).map((value, index) => <Skeleton key={index} height="40px" />)
```
ThuTrinh, line 233: you can use `R.range(0, 10).map((_, index) => <Skeleton key={index} height="40px" />)`

app/javascript/components/Reporting/User.jsx
```diff
@@ -0,0 +1,46 @@
...
+            )
+          }
+        </Body>
+      </Table>
+      {/* TODO: Pagination */}
+    </div>
+  )
+}
+
+export default Reporting
```
ThuTrinh, line 46: I feel like this components folder is being abused. IMO this folder should contain components that are reusable throughout the application. Page specific components should go into a components folder under the page's folder.
e.g.
```
components/
- Callout
- Calenda
- OfficeFilter
- etc.
pages/admin/Reporting/components
- ReportPicker.jsx
- ReportHeader.jsx
```

We should probably have a discussion about this at some point.

app/javascript/pages/admin/Reporting/index.jsx
samzx: That's a fair point, I'll remove this and add the options as they're implemented.

app/javascript/pages/admin/Reporting/index.jsx
samzx: Let's try this once we have the other options 😛

app/javascript/pages/admin/Reporting/index.jsx
samzx: Yes good point!

app/javascript/components/Reporting/User.jsx
samzx: I agree, I've moved the component local to the reporting directory. It did indeed feel weird reaching out to the components folder.

app/javascript/pages/admin/Reporting/User.jsx
```diff
@@ -0,0 +1,54 @@
...
+            {/* TODO: toggle sort by hours */}
+            <HeaderCell>{t('volunteer_portal.admin.tab.reporting.hours')}</HeaderCell>
+          </HeaderRow>
+        </Head>
+        <Body>
+          {
+            R.pipe(
+              R.slice(pageSize * (page - 1), pageSize * page),
+              R.sort((a, b) => b.hours - a.hours),
+            )(users).map((user, index) =>
```
ThuTrinh, line 37: I would move this out into a variable above the `return` statement. The benefit is separating data from display and decluttering the JSX part of the function.

app/javascript/pages/admin/Reporting/User.jsx
```diff
@@ -0,0 +1,54 @@
...
+                <Cell>{user.email}</Cell>
+                <Cell>{user.office.name}</Cell>
+                <Cell>{user.hours}</Cell>
+              </Row>
+            )
+          }
+        </Body>
+      </Table>
+      <div style={{ height: 16 }} />
+      <Pagination totalPages={Math.floor(users.length / pageSize) + (users.length % pageSize > 0 ? 1 : 0)} pagePadding={0} currentPage={page} onChange={setPage} />
```
ThuTrinh, line 49: same as before, I would move this logic into a variable above the `return` statement.

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+import UserReporting from './User'
+import ReportingQuery from './queries/userReportingQuery.gql'
+import { FilterContext, officeFilterValueLens } from '/context'
+import OfficeFilter from '../../../components/OfficeFilter'
+
+const AdminSection = styled.div`
+  display: flex;
+  justify-content: space-around;
+  padding-bottom: 50px;
```
ThuTrinh, line 29: should we be using spacing for this or was the intention for it to be a constant 50px?

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+`
+
+const ToolbarHeader = styled.div`
+  display: flex;
+  justify-content: space-between;
+  border-bottom: 1px solid ${({theme}) => theme.palette.grey["200"]};
+  padding-bottom: 16px;
```
ThuTrinh, line 36: should we be using spacing for this or was the intention for it to be a constant 16px?

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+
+    if (Object.prototype.hasOwnProperty.call(changes, 'selectedItem')) {
+      const nextSelectedItem = changes.selectedItem
+      setDropDownTempSelectedItem(nextSelectedItem)
+    }
+  }
+
+  const handleDropdownOnSelect = item => {
+    setReport(item)
+  }
```
ThuTrinh, line 132: do we need a variable for this? could we not just use `setReport ` on line 140?

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+  return (
+    <AdminSection>
+      <div style={{ width: "100%" }}>
+        <ToolbarHeader>
+          <Dropdown
+              isOpen={dropDownIsOpen}
+              onSelect={handleDropdownOnSelect}
+              onStateChange={handleDropdownStateChange}
+            >
+            <Field style={{ width: 200 }}>
```
ThuTrinh, line 143: why are we using inline styles?

app/javascript/pages/admin/Reporting/index.jsx
```diff
@@ -0,0 +1,208 @@
...
+            </div>
+            :
+            <UserReporting
+              users={R.propOr([], 'users', userData)}
+              startDate={dateRange.start}
+              endDate={dateRange.end}
+              onStartChange={changeStartDate}
+              onEndChange={changeEndDate}
+              setOctectStream={setOctectStream}
+            />
```
ThuTrinh, line 198: typically the loading display is handled by the component itself. In this case `UserReporting`, which exposes a prop `loading` and it will switch internally to show the stencil or content. 

The benefits are to increase conposability and reduce repetition:
1. you could move `UserReporting` anywhere and the loading display moves with it
2. you don't need to duplicate stencil code if you use the component in multiple places.

Also you remove the need for the nasty nested ternary operator.

ThuTrinh (approved):

app/javascript/components/StyleFoundation/Box.jsx
```diff
@@ -0,0 +1,21 @@
...
+  margin: ${props => toPixelIfDefined(props.margin)};
+  margin-left: ${props => toPixelIfDefined(props.marginL)};
+  margin-right: ${props => toPixelIfDefined(props.marginR)};
+  margin-top: ${props => toPixelIfDefined(props.marginT)};
+  margin-bottom: ${props => toPixelIfDefined(props.marginB)};
+`
+
+export const FlexBox = styled(MarginBox)`
+  display: flex;
+`
```
ThuTrinh, line 21: These three things are components in their own right. Typically we have them as separate files just under `/components`.
Also, why not just support margin inside Box? as it stands, FlexBox doesn't support height and width.

app/javascript/components/StyleFoundation/Box.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const toPixelIfDefined = (size) => !R.isNil(size) && `${size}px`
+
+export const Box = styled.div`
+  height: ${props => props.height};
+  width: ${props => props.width};
+`
+
+export const MarginBox = styled(Box)`
+  margin: ${props => toPixelIfDefined(props.margin)};
```
ThuTrinh, line 12: this is a little restrictive as it doesn't allow us to do `margin: 4px 6px`.

app/javascript/components/StyleFoundation/Box.jsx
```diff
@@ -0,0 +1,21 @@
+import styled from 'styled-components'
+import * as R from 'ramda'
+
+const toPixelIfDefined = (size) => !R.isNil(size) && `${size}px`
```
ThuTrinh, line 4: it might be more flexible to let the caller decide on the unit and default to `px` when not specified.

app/javascript/components/StyleFoundation/Box.jsx
```diff
@@ -0,0 +1,21 @@
...
+  height: ${props => props.height};
+  width: ${props => props.width};
+`
+
+export const MarginBox = styled(Box)`
+  margin: ${props => toPixelIfDefined(props.margin)};
+  margin-left: ${props => toPixelIfDefined(props.marginL)};
+  margin-right: ${props => toPixelIfDefined(props.marginR)};
+  margin-top: ${props => toPixelIfDefined(props.marginT)};
+  margin-bottom: ${props => toPixelIfDefined(props.marginB)};
```
ThuTrinh, line 16: it might be more flexible to let the caller decide on the unit and default to `px` when not specified.

app/javascript/components/StyleFoundation/Box.jsx
samzx: I understand now, thanks for the 🍐

Issue #294, opened 2020-01-14
Add Tags to reporting


samzx:
The reporting for admins currently doesn't display tags. We should display the tags so when admins export, they should be able to do analytics on this fields :)

samzx:
We also want:
 * - Individual events by event type
 * - Organized events by event type
 * - Individual events by tag
 * - Organized events by tag
 * - Individual events by organization
 * - Organized events by organization
 * - Individual events total
 * - Organized events total

PR #450, opened 2020-10-09
Individual and Organized event reports by event type.

samzx:
## Description
Individual and organized event reports by event type.

image

## References
https://github.com/zendesk/volunteer_portal/issues/294

## Implementation notes
- Currently just using active record `.execute` (untested) sql query as it will be quicker to get this out the door. It's the current query that we're using to generate reports (dataclips in heroku). We should convert to active record dsl to add tests in the near future.
- Added sensitive object (queries behind authentication)

## Risks
* low


kerrisim (approved):
Awesome 🎉

PR #471, opened 2020-11-04
patch manifest, I got it wrong before

samzx:
This is the correct way


anita-miller (approved):

samzx:
cc @ThuTrinh (this is the correct way)

samzx:
fixes #463

PR #474, opened 2020-11-18
Event tag reports

samzx:
## Description
Same as https://github.com/zendesk/volunteer_portal/pull/450 but for tags. Also fixed some bugs.

## Risks (if any)
* low

### DIFF
--- a/app/graphql/resolvers/tag_report_resolver.rb
+++ b/app/graphql/resolvers/tag_report_resolver.rb
@@ -38,7 +38,7 @@ module TagReportResolver
         (SELECT SUM(duration), tag_id
         from events LEFT JOIN event_tags ON (events.id=event_tags.event_id) LEFT JOIN signups ON signups.event_id = events.id
         WHERE \"events\".\"deleted_at\" IS NULL 
-        #{office_id == 'all' ? '' : "AND individual_events.office_id=#{office_id}"}
+        #{office_id == 'all' ? '' : "AND events.office_id=#{office_id}"}
         AND (starts_at > '#{Time.at(args[:after])}')
         AND (starts_at < '#{Time.at(args[:before])}')
         GROUP BY tag_id) as tagDuration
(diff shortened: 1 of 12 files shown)


anita-miller (approved):

app/graphql/resolvers/tag_report_resolver.rb
```diff
@@ -0,0 +1,56 @@
...
+        (SELECT SUM(duration), tag_id
+        from individual_events INNER JOIN individual_event_tags ON individual_events.id=individual_event_tags.individual_event_id
+        WHERE \"individual_events\".\"deleted_at\" IS NULL
+        #{office_id == 'all' ? '' : "AND individual_events.office_id=#{office_id}"}
+        AND \"individual_events\".\"status\" = 1
+        AND (date > '#{Time.at(args[:after]).utc}')
+        AND (date < '#{Time.at(args[:before]).utc}')
+        GROUP BY tag_id) as tagDuration
+      ON tagDuration.tag_id=tags.id
+      "
```
omgbob, line 19: You could use a heredoc for the query string. You won't have to escape all the quotes then.

```ruby
and_office = office_id == 'all' ? '' : "AND individual_events.office_id=#{office_id}"
sql <<-SQL
  SELECT "id", "name", "sum" from tags INNER JOIN
  (SELECT SUM(duration), tag_id
  from individual_events INNER JOIN individual_event_tags ON individual_events.id=individual_event_tags.individual_event_id
  WHERE "individual_events"."deleted_at" IS NULL
  #{and_office}
  AND "individual_events"."status" = 1
  AND (date > '#{Time.at(args[:after]).utc}')
  AND (date < '#{Time.at(args[:before]).utc}')
  GROUP BY tag_id) as tagDuration
ON tagDuration.tag_id=tags.id
SQL
```

omgbob (approved):

app/graphql/resolvers/tag_report_resolver.rb
samzx: nice, didn't know that, thanks!

PR #522, opened 2021-01-27
Upgrade application to rails 6.0.0

ThuTrinh:
Upgrade application to rails 6.0.0

## Description
- Updated rails gem and a few related dependency gems
- Upgrade application to rails 6.0.0 by running `rails app:update` 

I ran the application locally and all seems fine 🤞🏼 

- [ ] mtest test/graphql/loaders/association_loader_test.rb
- [ ] mtest test/graphql/loaders/event_hours_loader_test.rb
- [ ] mtest test/graphql/loaders/signup_count_loader_test.rb
- [ ] mtest test/graphql/resolvers/event_resolver_test.rb
- [ ] mtest test/graphql/resolvers/user_resolver_test.rb
- [ ] mtest test/graphql/resolvers/volunteer_resolver_test.rb (@kerrisim )
- [ ] mtest test/models/event_test.rb (@kayison )
- [ ] mtest test/models/organization_test.rb (@kayison )

## CCs
@zendesk/volunteer

@samzx @kerrisim and @kayison @craig-day - would you mind having a scan of the new configs files that came with rails 6.0? I am not sure they are all correct.

## Risks (if any)
* [HIGH] - rails upgrade to 6.0.0 with a bunch of changes to how things are configured. not sure if this works in stg or prod 😨


samzx (commented):
I had to do
```
mkdir tmp
mkdir tmp/pids
```
before starting, other than that, it seems to work fine!

Also looks like there's lots of tests to fix 😢 

Don't know much about the config files unfortunately, but they seem fine!

ThuTrinh:
I am not too sure about this theory since I am not well versed in Ruby or Rails, I will need help to validate my theory.

Since the only changes are framework and dependency version updates, it is likely that the test are failing due to some breaking change Rails 6 has imposed on how ActiveRecord scope works?

From what I can tell, it looks like the tests are failing due to unexpected result from filtering queries. It looks like we are using `scope` to apply the filtering.

This [thread](https://github.com/Shopify/graphql-batch/issues/77#issuecomment-356964474) suggests we should not use the `scope` argument of `ActiveRecord::Associations::Preloader#preload` and warns that future internal refactoring of rails could break this undocumented feature.

I also found these thread discussing fixes for scope chaining no longer leaking into the class level. I don't quite understand if its related, will need some assistance to dig deeper.
- https://github.com/rails/rails/pull/32380
- https://github.com/rails/rails/pull/35280

cc/ @kerrisim @kayison

ThuTrinh:
This is a massive PR with lots of change. Closing it in favour of smaller PRs that will address a smaller scope of the upgrade.

PR #525, opened 2021-01-29
Manually add nokogiri 1.11.1

kayison:
## Description
Build was failing due to nokogiri missing (the darwin version existing still caused it to fail). Tried a bunch of the usual bundle config changes to help install but that didn't work. Installed by downloading the gem manually from the ruby gems website and added it to the project https://rubygems.org/gems/nokogiri

## Risks (if any)
* Low: Build will fail if incorrect


samzx (approved):
🤞

PR #566, opened 2021-03-11
Create dependabot.yml

kerrisim:
## Description
Adding the dependabot.yml should reduce the frequency that dependabot checks for updates.
Monthly seems to the longest interval available according to the [docs.](https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#scheduleinterval)

## Risks (if any)
* [Low] - Adding configuration for dependabot


samzx (approved):
Nice! Thanks for this :) 

Seems like in the future they might let us do quarterly https://github.com/zendesk/volunteer_portal/pull/566 🙏

PR #604, opened 2021-06-30
add default organization as External Volunteering

samzx:
So when a user records an individual event, if the organization isn't listed in the list of organizations to select, they may choose this option.


samzx:
@ovillegaszen we'll need to re-run the translations

kerrisim (approved):

PR #614, opened 2021-07-14
Bump omniauth-saml

kennethc:
To fix incompatibility introduced by rexml-3.2.5.

## Description
When rexml was bumped in https://github.com/zendesk/volunteer_portal/pull/585, it introduced an incompatibility which broke SAML login. The issue was [fixed in the gem consuming rexml](https://github.com/onelogin/ruby-saml/issues/577). This PR bumps the direct dependency along that full chain to make sure that everything is up-to-date.

## Risks (if any)
* medium - can introduce more issues with saml login


samzx (approved):
Great work finding the issue 👏

PR #626, opened 2021-09-08
Set dependabot to only allow direct dependency upgrades

kennethc:
## Description
Dependabot is currently set to bump all dependencies. This could lead to a situation where only an indirect dependency is bumped, which introduces a breakage in a direct dependency, which can lead to breakages in the Portal. Refer to #614 for some more background.

This change limits dependabot to only upgrade direct dependencies.

## CCs

If app migration related
@zendesk/volunteer

## Risks (if any)
* Low - changes the behaviour of dependabot dependency upgrades

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,11 @@ updates:
     schedule:
       interval: "monthly"
     vendor: true
+    allow:
+      - dependency-type: "direct"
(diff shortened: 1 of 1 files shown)


kennethc:
I _think_ this will also include any required indirect dependency upgrades. At least it seems to be the case judging by existing PRs. The documentation is a little vague on this.

